PR #44, opened 2019-09-05
add all.json stat file

willr3:


core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsStore.java
```diff
@@ -84,12 +85,309 @@ public void record(String address, int phaseId, int stepId, String metric, Stati
...
+         }
+         if(!fork.equals(phaseSplit[2])) {//fork name should always be new
+            jGenerator.writeFieldName(phaseSplit[2]);
+            fork = phaseSplit[2];
+
+            jGenerator.writeStartObject();//start fork
+            jGenerator.writeFieldName(key );
+            jGenerator.writeStartObject();//all metric for fork
+         }
+         //jGenerator.writeFieldName(data.metric);
```
rvansa: forgotten?

core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsStore.java
```diff
@@ -99,6 +397,202 @@ public void persist(Path dir) throws IOException {
...
+      jGenerator.writeEndArray();//end failure array
+
+      //per phase.metric histogram and series
+      jGenerator.writeFieldName("phase");
+      phaseIterFor(jGenerator,sorted,(data)->data.phase,(key,data)->{
+         try {
+            jGenerator.writeFieldName(data.metric);
+            jGenerator.writeStartObject();//start metric
+
+            double outputValueUnitScalingRatio = 1000_000.0;
```
rvansa: constants should be static final fields

core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsStore.java
```diff
@@ -84,12 +85,309 @@ public void record(String address, int phaseId, int stepId, String metric, Stati
...
+
+         jGenerator.writeEndObject();
+      }
+      jGenerator.writeEndArray();//end total.csv
+   }
+   public <T> void phaseIterFor(JsonGenerator jGenerator,
+                                T[] sorted,
+                                Function<T,String> toString,
+                                BiConsumer<String,T> consumer,
+                                Consumer<String> onNewPhase,
```
rvansa: I wonder if it could look better had you moved all the lambdas to an object with multiple methods...

core/src/test/java/io/hyperfoil/core/impl/statistics/StatisticsStoreTest.java
```diff
@@ -0,0 +1,108 @@
...
+      snapshot.histogram.recordValue(1_000);
+      snapshot.histogram.recordValue(10_000);
+      snapshot.histogram.recordValue(30_000);
+      snapshot.histogram.recordValue(60_000);
+      snapshot.histogram.recordValue(120_000);
+      snapshot.histogram.recordValue(666_000_000);
+
+      store.record("address",phase.id(), step.id(), "metric1",snapshot);
+      store.recordSessionStats("address",1*60_000,"phaseName/001/test",1,10);
+//      snapshot.sequenceId=101;
```
rvansa: could you drop code that's commented out?

core/src/test/java/io/hyperfoil/core/impl/statistics/StatisticsStoreTest.java
```diff
@@ -0,0 +1,108 @@
...
+//      snapshot.histogram.setStartTimeStamp(2*60_000);
+//      snapshot.histogram.setEndTimeStamp(3*60_000);
+//      store.record("address",1,1,"metric1",snapshot);
+      snapshot.sequenceId=201;
```
rvansa: I still don't think it's a good idea to overwrite the snapshot object - the one you've recorded on line 66 is probably stored somewhere and you're modifying it (no defensive copy made).

core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsStore.java
```diff
@@ -84,12 +85,309 @@ public void record(String address, int phaseId, int stepId, String metric, Stati
...
+      for(Iterator<HistogramIterationValue> iter = iterationValues.iterator(); iter.hasNext();){
+         HistogramIterationValue iterValue = iter.next();
+
+         jGenerator.writeStartObject();
+         jGenerator.writeNumberField("Value",iterValue.getValueIteratedTo() / outputValueUnitScalingRatio);
+         jGenerator.writeNumberField("Percentile",iterValue.getPercentileLevelIteratedTo() / 100.0D);
+         jGenerator.writeNumberField("TotalCount",iterValue.getTotalCountToThisValue());
+
+         if(iterValue.getPercentileLevelIteratedTo() != 100.0D){
+            jGenerator.writeNumberField("1/(1-Percentile)",1.0D / (1.0D - iterValue.getPercentileLevelIteratedTo() / 100.0D));
```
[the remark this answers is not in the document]
willr3: We already have a dependency on direct floating point comparison...
https://github.com/HdrHistogram/HdrHistogram/blob/master/src/main/java/org/HdrHistogram/AbstractHistogram.java#L1801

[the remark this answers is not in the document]
rvansa: Ok, use whatever variant you prefer.

core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsStore.java
```diff
@@ -84,12 +85,309 @@ public void record(String address, int phaseId, int stepId, String metric, Stati
...
+         }
+         jGenerator.writeEndObject();
+      }
+      jGenerator.writeEndArray();//end histogram
+   }
+   public void seriesArray(JsonGenerator jGenerator, List<StatisticsSummary> series) throws IOException{
+      jGenerator.writeStartArray();//series
+      if(series!=null){
+         for(StatisticsSummary summary : series){
+            jGenerator.writeStartObject();
```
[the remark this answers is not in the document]
willr3: I used the same field names and flat structure as the csv files to minimize the changes. directly writing the summary will have different field names and the percentiles will be nested under `percentileResponseTime` 
directly writing the summary:
```json
{
  "startTime": 180000,
  "endTime": 240000,
  "minResponseTime": 1000,
  "meanResponseTime": 110836385,
  "maxResponseTime": 666894335,
  "meanSendTime": 0,
  "percentileResponseTime": {
    "50.0": 30079,
    "90.0": 666894335,
    "99.0": 666894335,
    "99.9": 666894335,
    "99.99": 666894335
  },
  "connectFailureCount": 0,
  "requestCount": 6,
  "responseCount": 0,
  "status_2xx": 0,
  "status_3xx": 0,
  "status_4xx": 0,
  "status_5xx": 0,
  "status_other": 0,
  "invalid": 0,
  "cacheHits": 0,
  "resetCount": 0,
  "timeouts": 0,
  "blockedCount": 0,
  "blockedTime": 0
}
```
manually writing the summary using the same field names as the csv
```json
{
  "Start": 180000,
  "End": 240000,
  "Requests": 6,
  "Responses": 0,
  "Mean": 110836385,
  "Min": 1000,
  "p50.0": 30079,
  "p90.0": 666894335,
  "p99.0": 666894335,
  "p99.9": 666894335,
  "p99.99": 666894335,
  "Max": 1000,
  "MeanSendTime": 0,
  "ConnFailure": 0,
  "Reset": 0,
  "Timeouts": 0,
  "2xx": 0,
  "3xx": 0,
  "4xx": 0,
  "5xx": 0,
  "Other": 0,
  "Invalid": 0,
  "CacheHits": 0,
  "BlockedCount": 0,
  "BlockedTime": 0
}
```

core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsStore.java
```diff
@@ -84,12 +85,309 @@ public void record(String address, int phaseId, int stepId, String metric, Stati
...
+   }
+   public void totalArray(JsonGenerator jGenerator, Data[] dataList,Function<Data,StatisticsSummary> getSummary, BiConsumer<JsonGenerator,Data> also) throws IOException{
+      jGenerator.writeStartArray();
+      for(Data data : dataList){
+         StatisticsSummary summary = getSummary.apply(data);//data.total.summary(percentiles);
+         if(summary==null){
+            continue;
+         }
+
+         jGenerator.writeStartObject();
```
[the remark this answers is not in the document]
willr3: I'm leery of directly writing the object because then we expose member field names as public API in the json. Are we comfortable with the member field names being public API?

core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsStore.java
```diff
@@ -84,12 +85,309 @@ public void record(String address, int phaseId, int stepId, String metric, Stati
...
+
+         if(also!=null){
+            also.accept(jGenerator,data);
+         }
+
+         jGenerator.writeEndObject();
+      }
+      jGenerator.writeEndArray();//end total.csv
+   }
+   public <T> void phaseIterFor(JsonGenerator jGenerator,
```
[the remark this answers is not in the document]
willr3: fixed

core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsStore.java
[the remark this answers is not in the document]
rvansa: I wouldn't make any compromises just to make it behaving in some 'legacy' way. Choose the best solution and then fix anything that got broken.

core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsStore.java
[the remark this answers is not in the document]
rvansa: Yes, we already expose `StatisticsSummary` as JSON through REST interface. The object already hosts some JSON annotations.

core/src/test/java/io/hyperfoil/core/impl/statistics/StatisticsStoreTest.java
willr3: You're right, snapshots were not being copied in the calls to mergeSnapshots(...) so I updated it to create a unique snapshot.

core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsStore.java
willr3: I thought about an interface with default methods and using anonymous instances. If we would rather a `PhaseIterForkWalker` then I can certainly refactor to use that instead

rvansa: Why anonymous instances? I would just have them in a class instead of a 200-line method.

willr3: Does this look better?
https://github.com/willr3/Hyperfoil/blob/all.json/core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsStore.java#L314

PR #50, opened 2019-10-10
add info to all.json and write it from ControllerVerticle

willr3:


core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsStore.java
```diff
@@ -289,6 +286,10 @@ default void onEndFork(String phaseName, String iterName, String forkName) {
    }
 
    public void writeJson(JsonGenerator jGenerator) throws IOException {
+      writeJson(jGenerator, true);
+   }
+
+   public void writeJson(JsonGenerator jGenerator, boolean standalone) throws IOException {
```
rvansa: I don't see the boolean being used.

willr3: good catch, that would have been a bug fix shortly after being able to test it in the lab.

PR #62, opened 2019-11-18
Histogram

willr3:
Fix issue where info.json was included in all.json using default encoding and not as formatted json.
Add a fixed response time increment histogram to percentile histogram we already record.


rvansa:
I've cherry-picked, removing the commented code, myself.

PR #75, opened 2019-12-06
Add a reference null check on method field.

whitingjr:
With this commit the log files generated are now 23M instead of 33M. Due to fewer NPE messages in the agent log file.


rvansa:
I am closing this for now, if you happen to encounter the problem (with more info, preferrably) please reopen.

PR #98, opened 2020-02-08
Implement original Wrk command, rename existing Wrk command to Wrk2

johnaohara:
- Rename wrk and wrk.sh to wrk2 and wrk2.sh
 - Implement equivilant to wrk command
 - Increase calibration pahse to 6s, gives time for driver to warm up

Current `wrk` command implements wrk2 threading model.  wrk2 uses the open-model to drive load, whereas the original wrk command uses the closed-model.  This means that Hyperfoil can not be a drop in replacement for wrk.

This PR renames the wrk command and shell script to wrk2 and wrk2.sh respectively. There is a new wrk command and wrk.sh shell script that emulates wrk and serves as a drop-in replacement for workloads that are currently using the original wrk load generator


rvansa:
Makes sense, thanks. Looking forward to see some results comparison with the original tools.

Btw., with ASL2 we don't need to include the license shebangs in every file.

PR #99, opened 2020-02-10
Minor changes to wrk/wk2

johnaohara:
- Issue with displaying help info defaulting to wrk for wrk and wrk2 command
 - Set default values for connections, duration and threads to values used by original wrk/wrk2
 - Disable periodic throughput statistics written to output buffer. re-enabled via '-p' command option

I have fixed one issue wrt to resolving the correct help info.  The other changes make the commands behaviour like the original wrk and wrk2 commands, so implements the same default values and requires a new cl option '-p' to enable periodic output of throughput during the test phase of the commands

### DIFF
--- a/cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
+++ b/cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
@@ -106,7 +106,7 @@ public abstract class WrkAbstract {
          }
          if (cr != null) {
             try {
-               System.out.println(cr.getCommandRegistry().getCommand("wrk", "wrk").printHelp("wrk"));
+               System.out.println(cr.getCommandRegistry().getCommand(getCommand(), getCommand()).printHelp(getCommand()));
             } catch (CommandNotFoundException ex) {
                throw new IllegalStateException(ex);
             }
@@ -117,13 +117,13 @@ public abstract class WrkAbstract {
 
    //   @CommandDefinition(name = "wrk", description = "Runs a workload simluation against one endpoint using the same vm")
    public abstract class AbstractWrkCommand implements Command<HyperfoilCommandInvocation> {
-      @Option(shortName = 'c', description = "Total number of HTTP connections to keep open", required = true)
+      @Option(shortName = 'c', description = "Total number of HTTP connections to keep open", defaultValue = "10")
       int connections;
 
-      @Option(shortName = 'd', description = "Duration of the test, e.g. 2s, 2m, 2h", required = true)
+      @Option(shortName = 'd', description = "Duration of the test, e.g. 2s, 2m, 2h", defaultValue = "10s")
       String duration;
 
-      @Option(shortName = 't', description = "Total number of threads to use.")
+      @Option(shortName = 't', description = "Total number of threads to use.", defaultValue = "2")
       int threads;
 
       @Option(shortName = 's', description = "!!!NOT SUPPORTED: LuaJIT script")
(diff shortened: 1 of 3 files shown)


cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
```diff
@@ -117,18 +117,21 @@ public void mainMethod(String[] args, Class wrkClass) {
...
       String duration;
 
-      @Option(shortName = 't', description = "Total number of threads to use.")
+      @Option(shortName = 't', description = "Total number of threads to use.", defaultValue = "2")
       int threads;
 
       @Option(shortName = 's', description = "!!!NOT SUPPORTED: LuaJIT script")
       String script;
 
+      @Option(shortName = 'p', description = "Periodic output of results during benchmark run", hasValue = false)
```
rvansa: Is this something specific to HF's implementation? If original wrk does not have it, we shouldn't either.

cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
```diff
@@ -117,18 +117,21 @@ public void mainMethod(String[] args, Class wrkClass) {
 
    //   @CommandDefinition(name = "wrk", description = "Runs a workload simluation against one endpoint using the same vm")
    public abstract class AbstractWrkCommand implements Command<HyperfoilCommandInvocation> {
-      @Option(shortName = 'c', description = "Total number of HTTP connections to keep open", required = true)
+      @Option(shortName = 'c', description = "Total number of HTTP connections to keep open", defaultValue = "10")
       int connections;
 
-      @Option(shortName = 'd', description = "Duration of the test, e.g. 2s, 2m, 2h", required = true)
+      @Option(shortName = 'd', description = "Duration of the test, e.g. 2s, 2m, 2h", defaultValue = "10s")
```
rvansa: Do both wrk and wrk2 have the same defaults?

cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
johnaohara: The original wrk did not write out the throughput every second.  So we either remove that completely or add a flag to enable it as additional behaviour. This flag restores the default behaviour

cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
johnaohara: yes

cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
rvansa: OK, I don't remember where this came from then. Please drop it completely.

johnaohara:
I have pushed an update with the periodic throughput reporting removed. Do you want me to squash commits?

rvansa:
Yes please. Btw. I think I've seen that you've added `wrk2.sh` at some point; this seems to be gone - is that intentional?

johnaohara:
wrk2.sh is still in the distribution src: https://github.com/johnaohara/Hyperfoil/blob/wrkTake2/distribution/src/main/resources/wrk2.sh

Is it missing from somewhere else?

johnaohara:
@rvansa have squashed a pushed

rvansa:
My bad, I forgot you had it in the other PR.

rvansa:
Integrated, thanks.

PR #107, opened 2020-04-16
use arrays for all.json and create legacy-all.json

willr3:


clustering/src/main/java/io/hyperfoil/clustering/ControllerVerticle.java
```diff
@@ -48,6 +50,8 @@
 import io.vertx.core.spi.cluster.NodeListener;
 import io.vertx.ext.cluster.infinispan.InfinispanClusterManager;
 
+import org.infinispan.commons.api.BasicCacheContainer;
```
rvansa, line 53: Really?

clustering/src/main/java/io/hyperfoil/clustering/ControllerVerticle.java
```diff
@@ -625,14 +629,52 @@ private void persistRun(Run run) {
             jGenerator.writeFieldName("info");
             jGenerator.writeRawValue(info.encode()); // writeObjectField() was encoding info as a POJO not json
 
-            JsonWriter.writeJson(run.statisticsStore, jGenerator, false);
+            jGenerator.writeFieldName("failures");
```
rvansa, line 632: Could we have this code in `JsonWriter`? Actually the whole code about creating the `all.json` would better fit there, just passing `info` as argument.

rvansa:
@willr3 Hi, are you going to update the PR?

clustering/src/main/java/io/hyperfoil/clustering/ControllerVerticle.java
willr3: What do you mean by "Really?"
The import is needed for shutdown() and I don't believe I created shutdown().

rvansa: Sorry, I did not notice that you've just reordered imports (I thought this was added by accident).

PR #136, opened 2020-12-23
Typo fix

gunnarmorling:
CC @stalep @johnaohara 

Should this be using https (instead of http), too?


rvansa:
Thanks for the fix.

Yes, we should use `https` in the example, though http returns 301 redirect which does not fail to demonstrate what we're doing here (and stats listing in https://hyperfoil.io/quickstart/quickstart1.html shows that correctly).

PR #144, opened 2021-01-21
Decode space with proper character

mossuchida:
Please review the code.  For HTTP2, it seems like it is simple string, I have use string methods to replace space.

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/client/netty/Http2Connection.java
+++ b/core/src/main/java/io/hyperfoil/core/client/netty/Http2Connection.java
@@ -27,6 +27,7 @@ import io.netty.handler.codec.http2.Http2Headers;
 import io.netty.handler.codec.http2.Http2Settings;
 import io.netty.util.collection.IntObjectHashMap;
 import io.netty.util.collection.IntObjectMap;
+import io.netty.util.internal.AppendableCharSequence;
 import io.hyperfoil.api.http.HttpResponseHandlers;
 import io.hyperfoil.api.session.Session;
 import io.vertx.core.logging.Logger;
@@ -130,6 +131,29 @@ class Http2Connection extends Http2EventAdapter implements HttpConnection {
       HttpClientPool httpClientPool = pool.clientPool();
 
       ByteBuf buf = bodyGenerator != null ? bodyGenerator.apply(request.session, this) : null;
+
+      if (request.path.contains(" ")) {
+          int length = request.path.length();
+          AppendableCharSequence temp = new AppendableCharSequence(length);
+          boolean beforeQuestion = true;
+          for (int i = 0; i < length; ++i) {
+              if (request.path.charAt(i) == ' ') {
+                  if (beforeQuestion) {
+                      temp.append('%');
+                      temp.append('2');
+                      temp.append('0');
+                  } else {
+                      temp.append('+');
+                  }
+              } else {
+                  if (request.path.charAt(i) == '?') {
+                      beforeQuestion = false;
+                  }
+                  temp.append(request.path.charAt(i));
+              }
+           }
+          request.path = temp.toString();
+      }
 
       Http2Headers headers = new DefaultHttp2Headers().method(request.method.name()).scheme(httpClientPool.scheme())
             .path(request.path).authority(httpClientPool.authority());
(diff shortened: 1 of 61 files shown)


core/src/main/java/io/hyperfoil/core/client/netty/Http2Connection.java
```diff
@@ -131,6 +131,19 @@ public void request(HttpRequest request,
 
       ByteBuf buf = bodyGenerator != null ? bodyGenerator.apply(request.session, this) : null;
 
+      if (request.path.contains(" ")) {
+          int question = request.path.indexOf("?");
+          String subFirst = "";
+          String subSecond = "";
+          if (question != -1) {
+              subFirst = request.path.substring(0 , question);
+              subFirst = subFirst.replaceAll(" ", "%20");
```
rvansa, line 140: I've suggested replacing that part with `AppendableCharSequence`, was there an issue with that? Using 2x substring, string concatenation and regexp search (!) through `replaceAll` seems rather excessive from memory allocation POV.

rvansa:
If you don't want to bother with HTTP2 you can leave it as it is in the PR, but please add a testcase for both HTTP1.1 and HTTP2.

core/src/main/java/io/hyperfoil/core/client/netty/Http2Connection.java
mossuchida: I will try to use AppendableCharSequence instead today.

mossuchida:
@rvansa I have updated HTTP2, also added DecodeTest, however, when I tried to package, mvn hangs at my test HttpDecodeSpaceTest. Would you have a look if it looks okay?  I wanted to test it without creating mock server, however, I could not figure out how to unit test, I copied HttpVersionsTest instead.

rvansa:
Thanks, I'll take a look.

rvansa:
Have you tried to run the test in an IDE or check the test logs? `core/target/surefire-reports/io.hyperfoil.core.client.HttpDecodeSpaceTest-output.txt` shows pretty clear message:
```
12:06:44,843 ERROR (vert.x-eventloop-thread-1) [i.v.c.i.ContextImpl] Unhandled exception java.lang.IllegalArgumentException: code: -1 (expected: 0+)
        at io.vertx.core.http.impl.Http2ServerResponseImpl.setStatusCode(Http2ServerResponseImpl.java:168)
        at io.hyperfoil.core.client.HttpDecodeSpaceTest.isPathCorrect(HttpDecodeSpaceTest.java:201)
```
(even the first `/ping` test tries to use status -1 which is illegal).

rvansa:
Btw when you fix the test please rebase & squash the commits; I've (hopefully) fixed the Github Actions so that it doesn't fail with PRs due to missing Quay credentials.

core/src/test/java/io/hyperfoil/core/client/HttpDecodeSpaceTest.java
```diff
@@ -0,0 +1,206 @@
...
+      } else {
+          int question = url.indexOf("?");
+          String subFirst = "";
+          String subSecond = "";
+          if (question != -1) {
+              subFirst = url.substring(0 , question);
+              subSecond = url.substring(url.lastIndexOf("?") + 1);
+              if (subFirst.contains("+")) {
+                  status = 601;
+              } else if (subFirst.contains("%20")) {
```
rvansa, line 186: This condition is superfluous (status becomes 200 anyway).

core/src/test/java/io/hyperfoil/core/client/HttpDecodeSpaceTest.java
```diff
@@ -0,0 +1,206 @@
...
+                  status = 200;
+              } else {
+                  status = 200;
+              }
+              if (status == 200) {
+                  if (subSecond.contains("%20")) {
+                      status = 602;
+                  } else if (subSecond.contains("+")) {
```
rvansa, line 194: Same as above.

core/src/test/java/io/hyperfoil/core/client/HttpDecodeSpaceTest.java
mossuchida: I have removed unnecessary if conditions.  Please review.

core/src/test/java/io/hyperfoil/core/client/HttpDecodeSpaceTest.java
mossuchida: I have removed unnecessary if conditions.  Please review.

PR #157, opened 2021-02-25
UUID Generator Support

diegolovison:


api/src/main/java/io/hyperfoil/function/SerializableToIntFunction.java
```diff
@@ -1,7 +1,6 @@
 package io.hyperfoil.function;
 
-import java.io.Serializable;
 import java.util.function.ToIntFunction;
 
-public interface SerializableToIntFunction<T> extends ToIntFunction<T>, Serializable {
+public interface SerializableToIntFunction<T> extends ToIntFunction<T>, SerializableToObjectFunction<T> {
```
rvansa, line 5: Classes in `i.h.function` are meant to be a straight extension of `java.util.function.*` with added `Serializable`. Besides, extending `SerializableToObjectFunction` here makes little sense.

api/src/main/java/io/hyperfoil/function/SerializableToObjectFunction.java
```diff
@@ -0,0 +1,6 @@
+package io.hyperfoil.function;
+
+import java.io.Serializable;
+
+public interface SerializableToObjectFunction<T> extends Serializable {
```
rvansa, line 5: This has no method and is not a function.

api/src/main/java/io/hyperfoil/function/ToStringFunction.java
```diff
@@ -0,0 +1,13 @@
...
+@FunctionalInterface
+public interface ToStringFunction<T> {
+
+    /**
+     * Applies this function to the given argument.
+     *
+     * @param value the function argument
+     * @return the function result
+     */
+    String applyAsString(T value);
```
rvansa, line 12: You don't need a special `ToStringFunction`, a regular `Function<String, T>` is fine.

core/src/main/java/io/hyperfoil/core/generators/ObjectValueProviderBuilder.java
```diff
@@ -0,0 +1,74 @@
+package io.hyperfoil.core.generators;
+
+import io.hyperfoil.api.config.BenchmarkDefinitionException;
+import io.hyperfoil.api.config.InitFromParam;
+import io.hyperfoil.api.session.Access;
+import io.hyperfoil.api.session.Session;
+import io.hyperfoil.core.session.SessionFactory;
+import io.hyperfoil.function.SerializableToObjectFunction;
+
+public abstract class ObjectValueProviderBuilder<P, V> implements InitFromParam<ObjectValueProviderBuilder<P, V>> {
```
rvansa, line 10: Ints are not Objects, so I'd rather call it `BaseValueProviderBuilder`.

core/src/main/java/io/hyperfoil/core/generators/ObjectValueProviderBuilder.java
```diff
@@ -0,0 +1,74 @@
...
+   public ObjectValueProviderBuilder<P, V> fromVar(String fromVar) {
+      this.fromVar = fromVar;
+      return this;
+   }
+
+   public abstract <R extends SerializableToObjectFunction<Session>> R buildFromAccess(Access access);
+
+   public abstract <R extends SerializableToObjectFunction<Session>> R buildFromConst();
+
+   public <R extends SerializableToObjectFunction<Session>> R build() {
```
rvansa, line 60: There shouldn't be any boxing/unboxing during runtime - in the `build()` itself it's fine but the concrete class must return either `SerializableToIntFunction<Session>` or `SerializableFunction<Session, String>`

core/src/main/java/io/hyperfoil/core/generators/RandomStringStep.java
```diff
@@ -0,0 +1,102 @@
...
+            throw new BenchmarkDefinitionException("Missing target var.");
+         }
+         if (stringGeneratorType == null) {
+            throw new BenchmarkDefinitionException("Missing target generator.");
+         }
+         return Collections.singletonList(new RandomStringStep(SessionFactory.access(toVar), stringGeneratorType));
+      }
+   }
+
+   private enum StringGeneratorType {
```
rvansa, line 93: What's the advantage in having
```
- randomStringGenerator: foo <- UUID
- randomStringGenerator: bar <- SomeOtherEnumValue
```
vs.
```
- randomUUID: foo
- randomSomethingElse: bar
```
The second variant seems more concise and can use code completion (IDE can suggest you randomUUID or randomSomethingElse, it would have hard time suggesting you enum values in a custom string).

core/src/test/java/io/hyperfoil/core/generators/RandomStringStepTest.java
```diff
@@ -0,0 +1,41 @@
...
+import io.hyperfoil.api.session.Session;
+import io.hyperfoil.core.session.SessionFactory;
+import io.hyperfoil.core.test.TestUtil;
+
+public class RandomStringStepTest {
+
+   @Test
+   public void testStringGenerator() {
+      // builder
+      String[] vars = new String[]{ "first", "second", "third", "fourth" };
```
rvansa, line 21: Does using 4 different var names test anything extra to using just single variable?
Btw. `var` is not the best name for a variable since it's a keyword (though positional to not break backwards compatibility) - it's confusing GitHub UI.

rvansa (changes requested):

api/src/main/java/io/hyperfoil/function/SerializableToIntFunction.java
diegolovison: Removed `SerializableToObjectFunction `

api/src/main/java/io/hyperfoil/function/SerializableToObjectFunction.java
diegolovison: Removed `SerializableToObjectFunction `

api/src/main/java/io/hyperfoil/function/ToStringFunction.java
diegolovison: Removed `ToStringFunction `

core/src/main/java/io/hyperfoil/core/generators/ObjectValueProviderBuilder.java
diegolovison: Renamed to `BaseValueProviderBuilder `

core/src/main/java/io/hyperfoil/core/generators/ObjectValueProviderBuilder.java
diegolovison: Those 3 methods were moded to the concrete implementation

core/src/main/java/io/hyperfoil/core/generators/RandomStringStep.java
diegolovison: Renamed to RandomUUIDStep

core/src/test/java/io/hyperfoil/core/generators/RandomStringStepTest.java
diegolovison: Test nothing extra but I liked the example with 4 diff names.
Renamed `var` to `varName`

core/src/main/java/io/hyperfoil/core/generators/RandomUUIDStep.java
```diff
@@ -0,0 +1,74 @@
...
+public class RandomUUIDStep implements Step, ResourceUtilizer {
+   private final Access toVar;
+
+   public RandomUUIDStep(Access toVar) {
+      this.toVar = toVar;
+   }
+
+   @Override
+   public boolean invoke(Session session) {
+      toVar.setObject(session, UUID.randomUUID().toString());
```
rvansa, line 29: Two issues here:
* `UUID.randomUUID()` uses secure random source and therefore it is unsuitable for performance tests.
* You are allocating an object just to dump it into string. Generate the `char[]` directly.

core/src/main/java/io/hyperfoil/core/generators/BaseValueProviderBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+   }
+
+   /**
+    * Initialize with a constant value.
+    *
+    * @param param Constant value.
+    * @return Self.
+    */
+   @Override
+   public abstract BaseValueProviderBuilder<P, V> init(String param);
```
rvansa, line 27: The abstract method is not necessary.

core/src/main/java/io/hyperfoil/core/generators/BaseValueProviderBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+
+   /**
+    * Initialize with a constant value.
+    *
+    * @param value Constant value.
+    * @return Self.
+    */
+   public BaseValueProviderBuilder<P, V> value(V value) {
```
rvansa, line 35: In order to support fluent API it would be better to return the actual self-type - see e.g. https://github.com/Hyperfoil/Hyperfoil/blob/3001e8ece3c82b06b44c08d582ced356a62e81ea/core/src/main/java/io/hyperfoil/core/builders/IntConditionBuilder.java#L38

core/src/main/java/io/hyperfoil/core/generators/StringValueProviderBuilder.java
```diff
@@ -0,0 +1,19 @@
+package io.hyperfoil.core.generators;
+
+public class StringValueProviderBuilder<P> extends BaseValueProviderBuilder<P, String> {
```
rvansa, line 3: Now I've noticed the class is not used at all - that renders most of the changes above unnecessary.

api/src/main/java/io/hyperfoil/function/SerializableToIntFunction.java
```diff
@@ -1,7 +1,6 @@
 package io.hyperfoil.function;
 
-import java.io.Serializable;
 import java.util.function.ToIntFunction;
 
-public interface SerializableToIntFunction<T> extends ToIntFunction<T>, Serializable {
+public interface SerializableToIntFunction<T> extends ToIntFunction<T> {
```
rvansa, line 5: You've lost the `Serializable` here

rvansa (changes requested):

core/src/main/java/io/hyperfoil/core/generators/RandomUUIDStep.java
rvansa: unsuitable = while you can generate millions of pseudo-random number per second the secure source can give you kilobytes per second of random data at best.

core/src/main/java/io/hyperfoil/core/generators/BaseValueProviderBuilder.java
diegolovison: Removed init

core/src/main/java/io/hyperfoil/core/generators/StringValueProviderBuilder.java
diegolovison: Can I keep the `BaseValueProviderBuilder` or should I rollback?

rvansa: Up to you; if you make it return correct self-type you can keep it. You should probably also inline the buildFromAccess and buildFromConst methods.

core/src/main/java/io/hyperfoil/core/generators/RandomUUIDStep.java
diegolovison: Created `LongFastUUID`

core/src/main/java/io/hyperfoil/core/generators/BaseValueProviderBuilder.java
diegolovison: Removed `BaseValueProviderBuilder `

core/src/main/java/io/hyperfoil/core/util/LongFastUUID.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Generate UUID based on the following match: [0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}
+ */
+public class LongFastUUID {
+   private LongFastUUID() {
+
+   }
+
+   private static final Long LONG_INSTANCE = Long.MAX_VALUE;
+   private static final Random RANDOM = ThreadLocalRandom.current();
```
rvansa, line 17: You must not store a thread-local instance in a global (static) variable.

core/src/main/java/io/hyperfoil/core/util/LongFastUUID.java
```diff
@@ -0,0 +1,35 @@
...
+         fastUUID = Long.class.getDeclaredMethod("fastUUID", long.class, long.class);
+      } catch (NoSuchMethodException e) {
+         throw new IllegalArgumentException(e);
+      }
+      fastUUID.setAccessible(true);
+   }
+
+   public static String randomUUID() {
+      try {
+         return (String) fastUUID.invoke(LONG_INSTANCE, RANDOM.nextLong(), RANDOM.nextLong());
```
rvansa, line 30: The correct pattern to invoking static methods via reflection is using `null` for the instance parameter.

rvansa (changes requested):

rvansa (approved):

rvansa:
Thanks!

PR #160, opened 2021-03-04
Fix Illegal reflective access LongFastUUID

diegolovison:
Tested with
```
openjdk version "17-ea" 2021-09-14
OpenJDK Runtime Environment (build 17-ea+12-906)
OpenJDK 64-Bit Server VM (build 17-ea+12-906, mixed mode, sharing)

```


rvansa:
Well I had in mind not doing stuff that JDK considers nasty but we can use this hack for a bit :)

diegolovison:
I was trying something like https://github.com/apache/groovy/pull/905/files but wasn't able to get it working.
Then, I decided to do the hack

PR #161, opened 2021-03-05
Hotrod client

diegolovison:


hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestBuilder.java
```diff
@@ -0,0 +1,59 @@
...
+   private HotRodOperationBuilder operation;
+   private HotRodCacheNameBuilder cacheName;
+   private int stepId = 1;
+
+   @Override
+   public List<Step> build() {
+
+      HotRodResourceKey key = new HotRodResourceKey();
+      HotRodRequestStep step = new HotRodRequestStep(stepId, key, operation.build(), cacheName.build(),
+            SessionFactory.access("cacheKey"), SessionFactory.access("cacheValue"));
```
rvansa, line 32: Don't hardcode variables for key and value. To align it with other similar properites in HF you should support these forms:
```yaml
key:
  value: foo
```
```yaml
key:
  fromVar: bar
```
```yaml
key:
  pattern: foo${bar}
```
```yaml
key: foo${bar} #inline pattern
```
Check out `StringGeneratorImplBuilder` (though you don't want to limit it to strings). The pattern (both inline and explicit) notation is optional, but `fromVar` and `value` (constant string) should be there.

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestStep.java
```diff
@@ -0,0 +1,98 @@
...
+
+      Object key = cacheKeyAccess.getObject(session);
+      Object value = cacheValueAccess.getObject(session);
+
+      HotRodDestinationTable destinations = HotRodDestinationTable.get(session);
+      HotRodConnectionPool pool = destinations.getPools().get(cacheName);
+      RemoteCache remoteCache = pool.getRemoteCache();
+
+      CompletableFuture future;
+      Statistics statistics = session.statistics(id(), HotRodMetric.METRIC_DURATION);
```
rvansa, line 62: `duration` is not a good name for metric - duration as for latency is just one aspect of the statistics.
Use session.currentSequence().definition().name() and delete the class `HotRodMetric`.

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestStep.java
```diff
@@ -0,0 +1,98 @@
...
+      long startTimestampMs = System.currentTimeMillis();
+      statistics.incrementRequests(startTimestampMs);
```
rvansa, line 64: You have incremented statistics before actually sending the request - move that after putAsync()

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestStep.java
```diff
@@ -0,0 +1,98 @@
...
+         session.proceed();
+      });
+      session.getResource(futureWrapperKey).set(future, startTimestampNanos, startTimestampMs);
+
+      return true;
+   }
+
+   @Override
+   public void reserve(Session session) {
+      session.declareResource(futureWrapperKey, io.hyperfoil.hotrod.resource.HotRodResource::new);
```
rvansa, line 85: fully qualified class name?

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodResponseStep.java
```diff
@@ -0,0 +1,24 @@
+package io.hyperfoil.hotrod.steps;
+
+import io.hyperfoil.api.session.Session;
+import io.hyperfoil.api.statistics.Statistics;
+import io.hyperfoil.core.steps.StatisticsStep;
+import io.hyperfoil.hotrod.resource.HotRodResource;
+import io.hyperfoil.hotrod.resource.HotRodResourceKey;
+
+public class HotRodResponseStep extends StatisticsStep {
```
rvansa, line 9: I guess this class does not need to be a `StatisticsStep` anymore as it does not record statistics/define SLA.

hotrod/src/test/java/io/hyperfoil/hotrod/HotRodTest.java
```diff
@@ -0,0 +1,68 @@
...
+      if (hotrodServers != null) {
+         hotrodServers.stop();
+      }
+   }
+
+   @Test
+   public void test() {
+      Properties properties = new Properties();
+      properties.put("hotRodPort", hotrodServers.getPort());
+      Benchmark benchmark = loadScenario("scenarios/HotRodTest.hf.yaml", properties);
```
rvansa, line 63: I've already mentioned that properties replacement is not acceptable. Please use override the loadScenario(...) in the test to change the port in the BenchmarkBuilder rather than replacing properties.

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestStep.java
```diff
@@ -0,0 +1,98 @@
...
+      Statistics statistics = session.statistics(id(), HotRodMetric.METRIC_DURATION);
+      long startTimestampMs = System.currentTimeMillis();
+      statistics.incrementRequests(startTimestampMs);
+      long startTimestampNanos = System.nanoTime();
+      if (HotRodOperation.PUT.equals(operation)) {
+         future = remoteCache.putAsync(key, value);
+      } else {
+         throw new IllegalArgumentException(String.format("HotRodOperation %s not implemented", operation));
+      }
+      future.thenRun(() -> {
```
rvansa, line 71: You're missing error handling - if the request fails the test will get stuck forever. In the error handler part record a failure and call `session.stop()` as the last statement.

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestStep.java
```diff
@@ -0,0 +1,98 @@
...
+   public boolean invoke(Session session) {
+
+      String cacheName = this.cacheName.apply(session);
+      HotRodOperation operation = this.operation.apply(session);
+
+      Object key = cacheKeyAccess.getObject(session);
+      Object value = cacheValueAccess.getObject(session);
+
+      HotRodDestinationTable destinations = HotRodDestinationTable.get(session);
+      HotRodConnectionPool pool = destinations.getPools().get(cacheName);
```
rvansa, line 58: You should not use `cacheName` here. You have allowed the parser to register several endpoints (`localhost:11222`) and created RCM for these, so you should probably use `localhost:11222` (from another configurable function) to select the connection pool (or `null` for default one).
If you intend to limit the implementation to single Infinispan cluster please do not let `hotrod` accept a sequence of endpoints.

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodCacheNameBuilder.java
```diff
@@ -0,0 +1,22 @@
+package io.hyperfoil.hotrod.steps;
+
+import io.hyperfoil.api.config.BuilderBase;
+import io.hyperfoil.api.session.Session;
+import io.hyperfoil.function.SerializableFunction;
+
+@FunctionalInterface
+public interface HotRodCacheNameBuilder extends BuilderBase<HotRodCacheNameBuilder> {
```
rvansa, line 8: You do not need to create a new builder to provide strings; use `StringGeneratorImplBuilder` (okay I admit the name is somewhat clunky).

rvansa (changes requested):

core/src/main/java/io/hyperfoil/core/session/SessionImpl.java
```diff
@@ -149,7 +149,14 @@ public Session declareObject(Object key) {
    }
 
    public Object getObject(Object key) {
-      return ((ObjectVar) requireSet(key)).get();
+      Var var = requireSet(key);
+      if (var instanceof ObjectVar) {
+         return ((ObjectVar) var).get();
+      } else if (var instanceof IntVar) {
+         return ((IntVar) var).get();
```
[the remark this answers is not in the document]
rvansa: I am trying to say that the step needs to be aware of the potential conversion - this is why even if it would keep the code simpler and prevent repetition of blocks like the above we won't push that from steps to the session.

core/src/test/java/io/hyperfoil/core/session/BaseScenarioTest.java
```diff
@@ -67,8 +72,8 @@ protected Benchmark loadScenario(String name) {
       }
    }
 
-   protected Benchmark loadBenchmark(InputStream config) throws IOException, ParserException {
-      return BenchmarkParser.instance().buildBenchmark(config, TestUtil.benchmarkData());
+   protected Benchmark loadBenchmark(InputStream config, Properties properties) throws IOException, ParserException {
```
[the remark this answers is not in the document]
diegolovison: Yes. After hours :)

core/src/main/java/io/hyperfoil/core/parser/BenchmarkParser.java
```diff
@@ -106,8 +108,23 @@ public BenchmarkBuilder builder(String source, BenchmarkData data) throws Parser
...
+
+   public Benchmark buildBenchmark(InputStream inputStream, BenchmarkData data, Properties properties)
+         throws ParserException, IOException {
+      String yaml = Util.toString(inputStream);
+      if (properties != null) {
+         Enumeration<String> enums = (Enumeration<String>) properties.propertyNames();
+         while (enums.hasMoreElements()) {
+            String key = enums.nextElement();
+            Object value = properties.get(key);
+            yaml = yaml.replaceAll(String.format("\\$\\{%s\\}", key), value.toString());
```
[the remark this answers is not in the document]
diegolovison: Ok. Used the same approach from `HttpScenarioTest`

core/src/test/java/io/hyperfoil/core/session/BaseScenarioTest.java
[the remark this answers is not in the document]
diegolovison: I decided to create `BaseScenarioTest::getScenarioReplacements`. In this case, I can reuse the same code.

hotrod/src/main/java/io/hyperfoil/hotrod/config/HotRodBuilder.java
```diff
@@ -0,0 +1,48 @@
...
+      this.cacheName = cacheName;
+      return this;
+   }
+
+   public HotRod build() {
+      HotRod hotRod = new HotRod(this.uri, this.cacheName);
+      return hotRod;
+   }
+
+   public String getUri() {
```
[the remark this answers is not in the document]
diegolovison: Removed `getUri()`

diegolovison:
The idea of this draft was to run the tests and compared them with what I have locally.
I appreciate your time review the PR :)

hotrod/src/main/java/io/hyperfoil/hotrod/connection/HotRodConnectionPoolImpl.java
```diff
@@ -0,0 +1,74 @@
...
+import io.hyperfoil.hotrod.api.HotRodConnectionPool;
+import io.netty.channel.EventLoop;
+import io.netty.channel.EventLoopGroup;
+
+public class HotRodConnectionPoolImpl implements HotRodConnectionPool {
+
+   private final HotRodClientPoolImpl hotRodClientPool;
+   private final EventLoop eventLoop;
+   private RemoteCacheManager remoteCacheManager;
+   private RemoteCache remoteCache;
```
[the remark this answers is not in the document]
diegolovison: I can't call `RCM.getCache(...)` because it will block.
I was thinking to allow multiple caches by `public void start(List<String> cacheNames)`, do the `RCM.getCache(...)` and store the result in a Map.
Also, `RemoteCache getRemoteCache()` should be replaced to do a get in the `Map`.
WDYT?

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestBuilder.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Issues a HotRod request and registers handlers for the response.
+ */
+@MetaInfServices(StepBuilder.class)
+@Name("hotrodRequest")
+public class HotRodRequestBuilder extends BaseStepBuilder<HotRodRequestBuilder> {
+
+   private HotRodOperationBuilder operation;
+   private HotRodCacheNameBuilder cacheName;
+   private int stepId = 1;
```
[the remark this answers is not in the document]
diegolovison: Done

core/src/main/java/io/hyperfoil/core/session/SessionImpl.java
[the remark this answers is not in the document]
diegolovison: The method is `Object getObject(Object key)`. Who is calling getObject is aware of potential boxing. Or my feeling is wrong ?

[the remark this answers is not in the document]
diegolovison: Also, `AsyncCache::CompletableFuture<V> putAsync(K key, V value)` will do the boxing even if I am doing a `put` with `int`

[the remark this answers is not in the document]
rvansa: Most people should not use the programmatic API but write the tests in YAML. When they create an integer and consume it as if it were an object they should get an error that will let them rethink if what they're doing is correct.

Yes, in your case the boxing is inevitable.

hotrod/src/main/java/io/hyperfoil/hotrod/connection/HotRodConnectionPoolImpl.java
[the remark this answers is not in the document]
rvansa: `RCM.getCache()` will block only when it's called the first time; afterwards it should just fetch the object from a map. So you need to provide a whitelist list of all caches that could be used, call `getCache()` when initializing the plugin, and then probably check in the pool if the `invoke()` is really accessing one of those white-listed caches. Hmm, such a whitelist is effectively the same as map... so ok, it makes sense to have your own map :)

core/src/main/java/io/hyperfoil/core/session/SessionImpl.java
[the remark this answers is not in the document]
diegolovison: Moved to `HotRodRequestStep`

hotrod/src/main/java/io/hyperfoil/hotrod/connection/HotRodConnectionPoolImpl.java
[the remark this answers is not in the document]
diegolovison: Renamed to `HotRodRemoteCachePool`

hotrod/src/main/java/io/hyperfoil/hotrod/resource/HotRodResourceKey.java
```diff
@@ -0,0 +1,7 @@
+package io.hyperfoil.hotrod.resource;
+
+import io.hyperfoil.api.session.Session;
+
+public class HotRodResourceKey implements Session.ResourceKey<HotRodResource> {
```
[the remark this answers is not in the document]
diegolovison: It is shared between `HotRodRequestStep` and `HotRodResponseStep`

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestBuilder.java
```diff
@@ -0,0 +1,59 @@
...
+   public List<Step> build() {
+
+      HotRodResourceKey key = new HotRodResourceKey();
+      HotRodRequestStep step = new HotRodRequestStep(stepId, key, operation.build(), cacheName.build(),
+            SessionFactory.access("cacheKey"), SessionFactory.access("cacheValue"));
+      HotRodResponseStep secondHotRodStep = new HotRodResponseStep(stepId, key);
+      return Arrays.asList(step, secondHotRodStep);
+   }
+
+   public HotRodRequestBuilder cacheName(String cacheName) {
```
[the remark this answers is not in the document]
diegolovison: What can I do in order to see the documentation locally?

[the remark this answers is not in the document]
rvansa: When you run `mvn install` it will create the markdown-docs in `distribution/target/steps/`. You can clone `https://github.com/Hyperfoil/hyperfoil.github.io` and copy contents of the generated folder into `docs/steps/` there and run `bundle exec jekyll serve` to build the web locally.

hotrod/src/main/java/io/hyperfoil/hotrod/resource/HotRodResourceKey.java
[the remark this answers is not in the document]
rvansa: Yes, `HotRodResource` is shared, so is the key. That doesn't prevent the `Key` to go inside `Resource` to keep related stuff together.
Might feel like nitpicking - it's ofc not an error to do it like you do, I am just trying to relay the idiosyncratic patterns.

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestBuilder.java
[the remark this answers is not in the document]
diegolovison: Under `distribution/target/steps/`, there is nothing related to `HotRod` even when I did
```
   /**
    * A named cache from the remote server if the cache has been defined, otherwise if the cache name is undefined,
    * it will return null.
    *
    * @param cacheName name of cache to retrieve
    * @return
    */
   public HotRodRequestBuilder cacheName(String cacheName) {
      return cacheName(() -> new HotRodCacheNameBuilder.Provided(cacheName));
   }
```

hotrod/src/main/java/io/hyperfoil/hotrod/connection/HotRodConnectionPoolImpl.java
```diff
@@ -0,0 +1,74 @@
...
+import org.infinispan.client.hotrod.RemoteCache;
+import org.infinispan.client.hotrod.RemoteCacheManager;
+import org.infinispan.client.hotrod.configuration.ConfigurationBuilder;
+import org.infinispan.client.hotrod.impl.transport.netty.ChannelFactory;
+
+import io.hyperfoil.hotrod.api.HotRodConnectionPool;
+import io.netty.channel.EventLoop;
+import io.netty.channel.EventLoopGroup;
+
+public class HotRodConnectionPoolImpl implements HotRodConnectionPool {
```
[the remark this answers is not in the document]
diegolovison: Removed or changed

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestStep.java
diegolovison: My new implementation is a cache per uri.

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestBuilder.java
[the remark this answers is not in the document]
rvansa: Sorry, you probably need to add hotrod here: https://github.com/Hyperfoil/Hyperfoil/blob/62329a6288436cde1d5e31181925801c7ba67008/distribution/pom.xml#L193

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestStep.java
diegolovison: Fixed

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestStep.java
diegolovison: Done

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestStep.java
diegolovison: Fixed

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodResponseStep.java
diegolovison: Done

hotrod/src/test/java/io/hyperfoil/hotrod/HotRodTest.java
diegolovison: Done

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestStep.java
diegolovison: Will `session.stop()` stops everything?
It is acceptable to have failures during a cache operation.
What `Statistics` method should I invoke to track errors?

rvansa: `session.stop()` will stop the current session - if you have more steps in the scenario these won't be executed. Executions of other sessions (virtual users) won't be affected. While we might think about multi-step scenarios where e.g. failure to insert the entry does not prevent other requests it's a sensible default to not invoke subsequent steps.

I shall revise the statistics to accomodate non-HTTP stuff better. For now please use `incrementResets` (this means an I/O error) and if that's some sort of TimeoutException use `incrementTimeouts`.

diegolovison: Done

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestBuilder.java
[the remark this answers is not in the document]
diegolovison: Done

hotrod/src/main/java/io/hyperfoil/hotrod/config/HotRodBuilder.java
```diff
@@ -0,0 +1,48 @@
...
+   }
+
+   // if needed, we can have/allow multiple caches using `,` as a split
+   public HotRodBuilder cacheName(String cacheName) {
+      this.cacheName = cacheName;
+      return this;
+   }
+
+   public HotRod build() {
+      HotRod hotRod = new HotRod(this.uri, this.cacheName);
```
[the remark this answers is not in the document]
diegolovison: What kind of code are you using for this kind of validation? For example: I can throw a null pointer if the value is null or empty but maybe you are using a different approach in Hyperfoil

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestBuilder.java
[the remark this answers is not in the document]
rvansa: The `put()` and `operation()` should get javadoc as well, especially that is what you expect your users to use.

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestStep.java
rvansa: Not done.
I) you're registering two handlers instead of one
II) `handle()` is invoked on success, too
III) `trackResponseError` is a no-op

hotrod/src/main/java/io/hyperfoil/hotrod/HotRodRunData.java
```diff
@@ -0,0 +1,82 @@
...
+         List<String> cacheNames = cacheNamesByUri.get(hotRod.getUri());
+         if (cacheNames == null) {
+            cacheNames = new ArrayList<>();
+         }
+         cacheNames.add(hotRod.getCacheName());
+         cacheNamesByUri.put(hotRod.getUri(), cacheNames);
+
+         // validation
+         if (allCaches.contains(hotRod.getCacheName())) {
+            throw new IllegalStateException(String.format("Duplicated cache: %s", hotRod.getCacheName()));
```
rvansa, line 42: TBH I can easily imagine a cross-site test where the cache names would be same...

hotrod/src/main/java/io/hyperfoil/hotrod/config/HotRod.java
```diff
@@ -0,0 +1,22 @@
+package io.hyperfoil.hotrod.config;
+
+import java.io.Serializable;
+
+public class HotRod implements Serializable {
+
+   private final String uri;
```
rvansa, line 7: So when you want to run a test with two caches in the same cluster you need to duplicate the endpoint part. Why did you do so? What is the advantage to having a list of caches nested under the endpoint?

I am worried that once you start adding configuration options (besides the URI), e.g. number of connections... you'll run into the fact that single RCM is shared by multiple caches (based on URI) but you could have conflicting configuration for the RCM in endpoint+cacheName pairs.

hotrod/src/main/java/io/hyperfoil/hotrod/config/HotRodBuilder.java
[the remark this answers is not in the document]
rvansa: `throw BenchmarkDefinitionException(...)`

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestBuilder.java
rvansa: A benchmark can write to multiple entries in different caches. You might want to execute those operations in parallel etc... When you do
```
- randomInt: 1 .. 100 -> someKey
- cachePut:
    key: ${someKey}
    value: ...
`
it's obvious that you've generated something and then you somehow use it.

File: hotrod/src/main/java/io/hyperfoil/hotrod/resource/HotRodResourceKey.java (hunk 16 above)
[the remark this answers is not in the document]
Comment (diegolovison) [reply hotrod/src/main/java/io/hyperfoil/hotrod/resource/HotRodResourceKey.java:5]: Commit: HotRodResource.Key

File: core/src/main/java/io/hyperfoil/core/metric/AuthorityAndPathMetric.java [hunk 22]
```diff
@@ -0,0 +1,9 @@
+package io.hyperfoil.core.metric;
+
+public class AuthorityAndPathMetric extends ProvidedMetricSelector {
```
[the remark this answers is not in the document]
diegolovison: Fixed by commit: MetricSelector

pom.xml
```diff
@@ -94,13 +95,15 @@
         <version.slf4j>1.7.12</version.slf4j>
         <version.snakeyaml>1.22</version.snakeyaml>
         <version.vertx>3.9.3</version.vertx>
+        <version.infinispan>12.0.1.Final</version.infinispan>
```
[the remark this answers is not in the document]
diegolovison: Fixed by commit: version.infinispan

hotrod/src/main/java/io/hyperfoil/hotrod/HotRodRunData.java
diegolovison: In this case, we should add the uri into the `hotrodRequest`. 
WDYT ?
```
name: hotrod-example
hotrod:
  uri: hotrod://localhost:11222
  cacheName: my-cache
usersPerSec: 1
duration: 5s
scenario:
  - example:
      - randomInt: cacheKey <- 1 .. 999
      - randomUUID: cacheValue
      - hotrodRequest:
          uri: hotrod://localhost:11222
          put: my-cache
```

hotrod/src/main/java/io/hyperfoil/hotrod/config/HotRod.java
diegolovison: How about adding `cacheNames` as a list under `hotrod`.
Or how about about replace `cacheName` ( single element ) with `cacheNames`
?

rvansa: I would do it this way:
```yaml
hotrod:
- uri: hotrod://site1:11222
  caches:
  - cacheA
  - cacheB
  connectionPool: ...
- uri: hotrod://site2:11222
  caches: [ cacheA ]
```

diegolovison: I didn't get why `connectionPool`. With the ISPN12 the `uri` is translated to `RCM`.

Example: `hotrod[s]://[username:password]@host[:port][,host[:port]…​][?property=value[&property=value…​]]`

rvansa: I meant `connectionPool` just as an example of additional configuration.
I didn't know that the properties can be passed in the query part; it would be nicer to have them exposed in a structured way as well.
If you use the URI as-is as the key in the map you'd probably have to repeat all the properties when setting `uri` for the individual request, which seems inconvenient and redundant.

diegolovison: I agree but could we improve that in another PR?

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestStep.java
diegolovison: Could you please double-check again?

hotrod/src/test/java/io/hyperfoil/hotrod/HotRodTest.java
```diff
@@ -0,0 +1,75 @@
...
+   @After
+   public void after(TestContext ctx) {
+      super.after(ctx);
+      if (hotrodServers != null) {
+         hotrodServers.stop();
+      }
+   }
+
+   @Test
+   public void test() {
```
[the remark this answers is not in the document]
diegolovison: Fixed by commit: HotRodFailuresTest

hotrod/src/main/java/io/hyperfoil/hotrod/config/HotRod.java
rvansa: Sure, I didn't try to imply that all the config properties must have mapping right now.

hotrod/src/main/java/io/hyperfoil/hotrod/HotRodRunData.java
```diff
@@ -0,0 +1,82 @@
...
+   private final HotRodPluginConfig plugin;
+   private HotRodRemoteCachePool[] pool;
+
+   public HotRodRunData(Benchmark benchmark, EventLoop[] executors, int agentId) {
+      this.plugin = benchmark.plugin(HotRodPluginConfig.class);
+
+      List<String> allCaches = new ArrayList<>();
+      Map<String, List<String>> cacheNamesByUri = new HashMap<>();
+      for (HotRod hotRod : this.plugin.hotRod()) {
+         List<String> cacheNames = cacheNamesByUri.get(hotRod.getUri());
```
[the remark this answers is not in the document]
diegolovison: Fixed by commit: computeIfAbsent

hotrod/src/main/java/io/hyperfoil/hotrod/HotRodRunData.java
diegolovison: Partially fixed by the commit: HotRod with multiple uri

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestBuilder.java
diegolovison: Can I have the scenario like?
```
scenario:
  - example:
      - randomInt: 1 .. 100 -> someKey
      - randomUUID: cacheValue
      - hotrodRequest:
          put: my-cache
          key: key-${someKey}-some-suffix
          value: ${cacheValue}
```

rvansa: Sure. However when you add the `get` operation you'll end up with
```yaml
- hotrodRequest:
    get: my-cache
    key: ...
    value: ??????
```
This construct is 'syntactically' correct (passes schema) but you have the throw an exception when the user provides a value for `get`. Then you add `putAll` with `keys` and in the end every operation will have to check presence of unrelated properties. That's why it would make sense to scope the properties for each operation, e.g.
```yaml
- hotrodRequest:
    cache: my-cache
    put:
      key: ...
      value: ...
```

diegolovison: Fixed with: Key and Value Generator

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodCacheNameBuilder.java
diegolovison: Fixed with: Key and Value Generator

hotrod/src/main/java/io/hyperfoil/hotrod/config/HotRodBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+
+   private String uri;
+   private String cacheName;
+
+   HotRodBuilder() {
+   }
+
+   @Override
+   public void readFrom(HotRodBuilder other) {
+
```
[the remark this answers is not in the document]
diegolovison: Fixed with: Clusters configuration & parsing

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestBuilder.java
```diff
@@ -0,0 +1,160 @@
...
+      PathMetricSelector selector = new PathMetricSelector();
+      this.metricSelector = selector;
+      return selector;
+   }
+
+   public StringGeneratorImplBuilder<HotRodRequestBuilder> cacheName() {
+      StringGeneratorImplBuilder<HotRodRequestBuilder> builder = new StringGeneratorImplBuilder<>(this, false);
+      cacheName(builder);
+      return builder;
+   }
```
rvansa, line 88: Seeing the missing empty lines I realized that the checkstyle is not running in this module, please add
```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-checkstyle-plugin</artifactId>
</plugin>
```
to the pom.xml

hotrod/src/main/java/io/hyperfoil/hotrod/steps/HotRodRequestBuilder.java
```diff
@@ -0,0 +1,160 @@
...
+
+   public StringGeneratorImplBuilder<HotRodRequestBuilder> key() {
+      StringGeneratorImplBuilder<HotRodRequestBuilder> builder = new StringGeneratorImplBuilder<>(this, false);
+      key(builder);
+      return builder;
+   }
+   public HotRodRequestBuilder key(StringGeneratorBuilder builder) {
+      if (this.key != null) {
+         throw new BenchmarkDefinitionException("Path generator already set.");
```
rvansa, line 116: I guess you meant `key already set`.

diegolovison:
Done.

rvansa (approved):

rvansa:
Thanks, it's quite a feat to get this implemented! I hope that in future PRs you'll add support for all other operations and maybe rethink how the `hotrodRequest` is configured. Since we're at 0.x versions there's no issue with backwards compatibility.

PR #165, opened 2021-03-19
Display user friendly exception message when selecting random item from empty collection

johnaohara:
Log a more user friendly exception message when `randomItem` tries to select from an empty collection;

message has changed from 
```
java.lang.IllegalStateException: bound must be positive
```

to

```
java.lang.IllegalStateException: Collection items is empty, can not select a random item
```


rvansa:
Thanks, integrated.

PR #202, opened 2021-10-15
Add support for loading controller local benchmarks

stephenc:
- This can be used as a pathway to have a configmap mounted benchmark easily accessible for automation

### DIFF
--- a/cli/src/main/java/io/hyperfoil/client/RestClient.java
+++ b/cli/src/main/java/io/hyperfoil/client/RestClient.java
@@ -158,20 +158,7 @@ public class RestClient implements Client, Closeable {
                }
                request.sendMultipartForm(multipart, handler);
             }, 0,
-            response -> {
-               if (response.statusCode() == 204) {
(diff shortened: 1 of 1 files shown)


stephenc:
If I mount the whole configmap as a directory then all the files will be changed at the same time, so again when I have multiple benchmarks I won't know the order in which they will be loaded and therefore I will not know which will be run with the `run` command

rvansa:
Ouch, I think where you got the confusion from :) So for your convenience the CLI remembers last benchmark you've worked with, and therefore the `run` will start that very benchmark. It will save you from typing its name, but in general you can use `run foobar` to run benchmark `foobar` and run that one. When you look at the server API it requires run name.
The notion of 'last benchmark' is on the client side only, therefore if some benchmarks change magically and you type `run` the CLI will still run the last benchmark you've worked with.
(there's a bit of difference with `last` for runId where this is understood server-wise - we don't need to keep it that way but it's convenient for my occassional manual `curl`).

Issue #222, opened 2022-04-21
benchmark template issues


vitorz:
Hello,
i guess there is a bug related to benchmark templates. The issue happens when you use "advanced" template features like `!foreach`.
Below a template quite similar to the one i'm using, this version works:
```name: cluster
agents: !foreach
  items: !param NODES
  do:
    !concat [ "hf-agent-", !param ITEM _ ]:
      node: !param ITEM _
      log: hf-agent-log/log4j2-agent.xml
http:
  host: http://some.url/context
  sharedConnections: 4000
phases:
  - main:
      always:
        users: 2000
        duration: 60s
        scenario:
          - test:
            - httpRequest:
               POST: /customers/PACKET_A
               sync: true
```
The above template works fine with parameter like this `-PNODES=myfirst-openshift-node,mysecond-openshift-node`
If in the above working template i add a new variable, for example replacing `http://some.url/context` with `!param TARGET_ROUTE` i got errors. Tried also to add default values with no success.

vitorz:
Hello,
i investigate a little more, the issue happens on "run" calls. In this example we have 2 parameters. The client send both parameters, i sniffed the traffic, but the ApiRouter parse only the first one, ignoring all the rest. It was a fast debug session, so i could be wrong. So the test to do is executing a "run" call with more than 1 parameter. About the template engine, yes probably it works fine and the problem is not there.

rvansa:
You're right, the problem is in ApiRouter: the generated code does this:
```
String _templateParam = queryParam(ctx, "templateParam", null);
List<String> templateParam = convert(_templateParam, List.class);
```
and the `queryParam(...)` fetches only the first instance of that parameter, while we use it multiple times. The `convert` then just creates a singleton list.

Would you care to write a fix, or should I try to find some time for that next week?

PR #223, opened 2022-04-26
fixing benchmark template with more than 1 param

vitorz:
This update should fix this [issue](https://github.com/Hyperfoil/Hyperfoil/issues/222). This change fits well the actual version of  openapi.yaml. Extending for future needs should be easy.


rvansa:
Thanks, great catch and fix!

PR #235, opened 2022-07-25
No Download messages while running mvn

diegolovison:


rvansa:
Hi @diegolovison was this created just to debug something in the CI? Shall I close it?

diegolovison:
imho, we should merge.
I don't like downloading messages and `-Dagent.log.trace` will create logs in the agents, IRRC, when something crashes in the agent.

rvansa:
I am not so keen on running with trace logging by default, but heck, this is CI, let's do it.

PR #245, opened 2022-09-29
Port should be optional when setting authority in the httpRequest step

afalhambra:


afalhambra:
@rvansa could you please review? thank you!

http/src/main/java/io/hyperfoil/http/config/HttpBuilder.java
```diff
@@ -98,6 +98,16 @@ public String host() {
       return host;
    }
 
+   public int port() {
+      if (port != -1) {
+         return port;
+      } else if (protocol != null) {
+         return protocol.portOrDefault(port);
+      } else {
+         return Protocol.fromPort(port).portOrDefault(port);
```
rvansa, line 107: I think that if `port == -1 && protocol == null` this method should rather throw a BenchmarkDefininitionException

http/src/main/java/io/hyperfoil/http/config/HttpBuilder.java
```diff
@@ -79,10 +79,10 @@ String name() {
    String authority() {
       if (host == null) {
          return null;
-      } else if (protocol != null) {
-         return host + ":" + protocol.portOrDefault(port);
+      } else if (port == -1) {
```
rvansa, line 82: I don't like that this method could return something ambiguous. Since you've added `port()`, could you inline the usages to make the semantics about port selection clear where it is used?

rvansa (changes requested):
Thanks for the tests, but let's yet clean up the semantics in HttpBuilder.

http/src/main/java/io/hyperfoil/http/config/HttpBuilder.java
rvansa: (because in this situation it can unexpectedly fall back to port 80)

afalhambra: ok, done

http/src/main/java/io/hyperfoil/http/config/HttpBuilder.java
afalhambra: done

rvansa (approved):
Thanks, LGTM

PR #246, opened 2022-10-03
No connection pool when authority and host are both defined without port

afalhambra:
Due to https://github.com/Hyperfoil/Hyperfoil/pull/245, running a benchmark fails at runtime when authority and host are defined without any port, as it complains that no authority is defined for a particular host


afalhambra:
@rvansa could you please have a look? thanks!

rvansa:
Integrating, thanks.

PR #266, opened 2023-01-17
Validate wrk/wrk2 connections and fail fast in case of error - Fixes #265

### LINKED ISSUE #265: Wrk/Wrk2 commands do not test remote URL before starting benchmark
The `wrk` or `wrk2` commands do not test the remote connection before starting the benchmark and these commands fail with a unclear error message;

```
Running 10s test @ http://localhost:8080/hello
  2 threads and 10 connections
java.lang.IllegalStateException: Missing stats for phase 'test'
```

these command should fail fast with the following error;

```
unable to connect to localhost:8080 Connection refused
```

johnaohara:


jesperpedersen:
Merged.

PR #273, opened 2023-02-04
Improved HTTP 1x decoding

franz1981:
Still draft given that's WIP; already opened to suggestions/questions/crititics

General description is:
1. `readHeaders` often pop up on cpu profiling sessions due to highly variable line length, causing bad misprediction(s) of parser/decoder next state, while reading byte-per-byte incoming data
2. `readHeaders`  contains `byte[]/String` heavy allocators (for header names/values)

The first has been solved by using https://github.com/netty/netty/pull/10737 I've introduced some time ago in Netty (and that show larger benefits as the number of bytes processed grows and when inputs sequences makes branch predictions useless), that's still better then any `ByteBuf::getByte`, which perform accessibility/refCnt and bound checks per-call, while `indexOf` perfom it just once.

The second has been solved by using already "latin" encoded strings available in Netty ie `AsciiString` which allows:
- saving an intermediate `byte[]` buffer to be used to create the `String`
- optimized `AsciiString::contentEqualsIgnoreCase` eg used on `HttpCacheImpl::responseHeader`,`HttpResponseHandlersImpl::handleHeader`
- optimized accesses by reaching the internal byte[], if necessary (for further improvements)

Some potential drawbacks of using `AsciiString` is that we won't have any JVM intrinsics available on it for some specific methods

### DIFF
--- a/api/src/main/java/io/hyperfoil/impl/Util.java
+++ b/api/src/main/java/io/hyperfoil/impl/Util.java
@@ -7,6 +7,7 @@ import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.nio.CharBuffer;
 import java.nio.charset.CharsetEncoder;
 import java.nio.charset.CoderResult;
@@ -23,6 +24,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import io.hyperfoil.api.config.Benchmark;
 import io.hyperfoil.api.config.BenchmarkDefinitionException;
 import io.netty.buffer.ByteBuf;
+import io.netty.util.AsciiString;
 
 public class Util {
    public static final CompletableFuture<Void> COMPLETED_VOID_FUTURE = CompletableFuture.completedFuture(null);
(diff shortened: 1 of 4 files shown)


jesperpedersen:
https://www.jmix.io/blog/utf-8-in-http-headers/

franz1981:
Thanks @jesperpedersen for https://www.jmix.io/blog/utf-8-in-http-headers/ so I see that, although few/many/all browsers will be decently able to use UTF-8 as header values (not names anyway, but in theory, is the same no? a malicious SW can almost send everything...) I could check if It's UTF-8.
But I don't think is the right way to do it; or we will have to add the many nuance of http validations (making us a copy of Netty http encoder, at this point, so why maintain our own?).
TBH even Netty does ingest bytes as latin data and verify just few few chars to be right, but won't try to detect UTF-8 encoding because is rare and malicious and not just out of the spec (as happen frequently with HTTP 1.x, sadly).

This is the usual performance/flexibility tradeoff, but if we feel safer by adding such check and slow path I can do it.

jesperpedersen:
We have to investigate if this pull request provides a performance benefit if all HTTP headers are indeed ISO-LATIN-1.

If so, what is the overhead of doing the headers check - per application or per request.

Think about going in another way - telling the workload that it can use ISO-LATIN-1 for everything, but keeping the default USF-8.

api/src/main/java/io/hyperfoil/impl/Util.java
```diff
@@ -89,6 +90,12 @@ public static String toString(ByteBuf buf, int offset, int length) {
       }
    }
 
+   public static AsciiString toAsciiString(ByteBuf buf, int offset, int length) {
+      final byte[] bytes = new byte[length];
```
jesperpedersen, line 94: I'm not sure about memory allocations in `static` methods. In most cases a preallocated memory segment can be used, and it can be "memset(asciiBuffer, 0, ...)" if necessary.

Since `length` is a parameter here - we could check against "sizeof(asciiBuffer)" and only allocate if needed.

malloc(), memset() and free() has a massive overhead in high performance application, and for Java that means GC as well. So, the fewer we can do the better

franz1981: the `static` context shouldn't change the mechanism used by JIT to perform allocations, but just the so-called "escape context": as long as something "escape" (meaning that the ref of allocated memory go out of scope to be used elsewhere) a real allocation is required, which magnitude depends by few JVM parameters (alignments, COOPS, etc etc) and where (TLAB/yng gen or old gen).

franz1981:
> if this pull request provides a performance benefit if all HTTP headers are indeed ISO-LATIN-1

Given that's the common use case I can run a CPU bound case i.e. that's stressing response decoding, for the most
Ideally I can pack a JMH benchmark too and add it as module to the project, to be used for other parts, wdyt?

api/src/main/java/io/hyperfoil/impl/Util.java
jesperpedersen: Yeah, but what is the improvement of using 

```
private static byte[] ASCII_BUFFER = new byte[ASCII_BUFFER_SIZE];
```

on a class level instead ?

jesperpedersen:
I think it would of benefit to have JMH in place for the project.

But maybe we should target this pull request only for master, and mark it as a breaking change in the release.

api/src/main/java/io/hyperfoil/impl/Util.java
franz1981: Class level one require synchronization, but in this specific case there are no temporary copies (that's the reason of the change).

In the original code a tmp `byte[]` was allocated and then passed to the JDK lib's `String`, that will internally use it to produce the internal `byte[]/char[]` holding the content of the `String`, while here we create a single `byte[]` and  passing it into a freshly (empty) `AsciiString` to be used as content (without any copy).

In total we had:
- one byte[] + one String (with it's internal byte[]/char[])

while here:
- one byte[] + one AsciiString

we're both saving a useless copy and allocation

jesperpedersen: Lets target `master` only

franz1981: I don't have labelling powers yet

franz1981:
@jesperpedersen I've added d8caf9a1ddcad1f6b2a3970354497e5a6e32aeb6 and opened https://github.com/netty/netty/pull/13197 to track the Netty issue (sigh!!): it's specular to the response decoding i.e. 
encoded headers just use ISO encoding to speed up. I can make uses of UTF-8 methods here too, but given that we're in control of what we send to the server, I prefer to be wide and within the spec, as said.

We can discuss and make this a toggle too and have 2 separate methods to capture that; I've pushed the commit to see in a quick test how much is different (I can tell already - it is)

franz1981:
@jesperpedersen Radim "knows" -> https://github.com/Hyperfoil/Hyperfoil/blob/e2ae9831cc356d03c379ad0473d954c59c693786/http/src/main/java/io/hyperfoil/http/connection/Http1xConnection.java#L164

franz1981:
@jesperpedersen 
The latest commit I've sent has break something; I'll keep on working on this on monday, but the idea is that I'm:
- forcing ASCII header names (that's perfectly fine and not a breaking change)
- checking header values if are latins, to speedup both encoding/decoding, falling back to UTF-8 in case they're not

the later has been implemented to save a breaking change to happen, at the cost of some batchy check in the hot path.
A performance evaluation will be done on monday while I fix the last commit sent.

franz1981:
I'm getting a test failure I cannot reproduce locally - @jesperpedersen can you reproduce it locally? any idea what's the problem?

franz1981:
> But, that isn't the issue - we need to figure out why it is failing. Can we use Byteman for this ?

I don't know - I see that such tests were failing in other prs hence are not related my changes here, but they don't look to be reproducible locally.
Please create an issue and I can investigate on the tests failure as a proper task (by understanding what the tests are supposed to do as well)

jesperpedersen:
Rebased, and merged.

PR #294, opened 2023-05-09
Fix RandomCsvRowStepTest popping out locator

lampajr:
Hello team, while running **all** tests from command line I faced an issue that I'll try to summarize below.

**Issue**

```bash
[ERROR] Failures: 
[ERROR]   InvalidBenchmarkTest.testDeadlock 
Expected: exception with message a string containing "Phase dependencies contain cycle"
     but: message was null
Stacktrace was: java.lang.UnsupportedOperationException
	at io.hyperfoil.core.test.TestUtil$2.locationMessage(TestUtil.java:43)
	at io.hyperfoil.api.config.BenchmarkDefinitionException.getMessage(BenchmarkDefinitionException.java:32)
	at io.hyperfoil.api.config.BenchmarkDefinitionException.<init>(BenchmarkDefinitionException.java:23)
```

and 

```bash
[ERROR]   InvalidBenchmarkTest.testMissingPhase 
Expected: exception with message a string containing " is not defined"
     but: message was null
Stacktrace was: java.lang.UnsupportedOperationException
	at io.hyperfoil.core.test.TestUtil$2.locationMessage(TestUtil.java:43)
	at io.hyperfoil.api.config.BenchmarkDefinitionException.getMessage(BenchmarkDefinitionException.java:32)
```

The issue was not appearing if running just those two tests only.
After some investigation it looks like that running `InvalidBenchmarkTest` and `RandomCsvRowStepTest` simultaneously reproduces the error, e.g., by running `mvn test -fn -pl core -Dtest=InvalidBenchmarkTest,RandomCsvRowStepTest`.

Given that I think I was facing a test concurrency issue where somehow the exception thrown by `RandomCsvRowStepTest#testParseAllColumns` is retrieved during the `InvalidBenchmarkTest` execution and if I got it properly it should be caused by the fact that in `testParseAllColumns` the Locator is not properly popped out (as done for all other tests).

> **NOTE**: Running all tests from IDE (e.g., IntellijIDEA) I was not able to reproduce exactly the same issue on `InvalidBenchmarkTest` but some others failed a reason that could be traced back to the same - after this fix also those tests seem fixed as well

**Tools**

```bash
❯ mvn --version
Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)
Maven home: /home/alampare/.sdkman/candidates/maven/current
Java version: 11.0.16, vendor: Eclipse Adoptium, runtime: /home/alampare/.sdkman/candidates/java/11.0.16-tem
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "6.2.14-200.fc37.x86_64", arch: "amd64", family: "unix"
```

### DIFF
--- a/core/src/test/java/io/hyperfoil/core/generators/RandomCsvRowStepTest.java
+++ b/core/src/test/java/io/hyperfoil/core/generators/RandomCsvRowStepTest.java
@@ -41,6 +41,7 @@ public class RandomCsvRowStepTest {
          cols.accept(pos, pos);
       }
       String[][] rows = ((RandomCsvRowStep) builder.build().get(0)).rows();
+      Locator.pop();
       Assert.assertEquals(rows.length, DATA.length);
       for (int i = 0; i < DATA.length; i++) {
          Assert.assertArrayEquals(DATA[i], rows[i]);


franz1981:
thanks @lampajr well done!

franz1981 (approved):
LGTM, thanks @lampajr !!

jesperpedersen:
Merged.

Issue #239, opened 2022-08-23
startWith with delay


rvansa:
For simulation spikey loads it might be useful to execute steady state and then add additional load in parallel. This could be defined through a combination of phase name and delay like
```
startWith:
  phase: anotherPhase
  delay: 120s
```
Combining that with other `startAfter`/`startAfterStrict` constraints or `startTime` should be illegal; we want these phases in sync.

lampajr:
Hey @jesperpedersen , is this still a valid issue? if so, may I try to take it?

jesperpedersen:
Sync with @franz1981 on it

lampajr:
great, thanks a lot @jesperpedersen !

PR #298, opened 2023-05-31
Additional startWith with delay feature

lampajr:
Fixes https://github.com/Hyperfoil/Hyperfoil/issues/239

**Improvement:**
With this new `startWith` with `delay` feature, users can actually sync phases between each other.

**Usage:**

This is an example where you could start a `phase1` after 5 seconds that the coupled `phase2` has started.

```yaml
name: first-swd
http:
  host: http://localhost:8080
  sharedConnections: 10
phases:
- phase1:
    constantRate:
      usersPerSec: 10
      duration: 10s
      startWith:
        phase: phase2
        delay: 5s
      scenario:
      - fetchIndex:
        - httpRequest:
            GET: /
- phase2:
    constantRate:
      usersPerSec: 10
      duration: 10s
      scenario:
      - fetchDetails:
        - httpRequest:
            GET: /offering/1

```

Its definition is quite similar to the `startAfter` feature, the main differences are:
- There is the additional `delay` field
- You **cannot** provide a list of `startWith` as it does not make sense by design, and of course circular dependencies are not allowed too.

> **Note:** right now if you set `delay=0` the code checks if the coupled phase has been already started before marking the current one as available. This means that the current phase will start in the next available slot, therefore even with `delay=0` there is an intrisic delay `> 0`

### DIFF
--- a/api/src/main/java/io/hyperfoil/api/config/BenchmarkBuilder.java
+++ b/api/src/main/java/io/hyperfoil/api/config/BenchmarkBuilder.java
@@ -30,7 +30,6 @@ import java.util.Queue;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import io.hyperfoil.impl.FutureSupplier;
 
@@ -147,9 +146,14 @@ public class BenchmarkBuilder {
       Map<String, Phase> phases = phaseBuilders.values().stream()
             .flatMap(builder -> builder.build(bs, phaseIdCounter).stream()).collect(Collectors.toMap(Phase::name, p -> p));
       for (Phase phase : phases.values()) {
+         // check if referenced dependencies exist
          checkDependencies(phase, phase.startAfter, phases);
          checkDependencies(phase, phase.startAfterStrict, phases);
          checkDependencies(phase, phase.terminateAfterStrict, phases);
+         if (phase.startWithDelay != null) {
+            checkDependencies(phase, Collections.singletonList(phase.startWithDelay.phase), phases);
+         }
+         checkStartWith(phase);
          checkDeadlock(phase, phases);
       }
       Map<String, Object> tags = new HashMap<>();
@@ -180,7 +184,9 @@ public class BenchmarkBuilder {
       toProcess.add(phase);
       while (!toProcess.isEmpty()) {
          Phase p = toProcess.poll();
-         Stream.concat(p.startAfter.stream(), p.startAfterStrict.stream()).forEach(name -> {
+         // consider all referenced dependencies (startAfter, startAfterStrict and startWithDelay)
+         // ensure there are no cyclic references.
+         p.getDependencies().forEach(name -> {
             Phase p2 = phases.get(name);
             if (p2 == null) {
                // non-existent phase, will be reported later
@@ -210,6 +216,17 @@ public class BenchmarkBuilder {
                   .map(name -> " Did you mean " + name + "?").orElse("");
             throw new BenchmarkDefinitionException("Phase " + dep + " referenced from " + phase.name() + " is not defined." + suggestion);
          }
+      }
+   }
+
+   /**
+    * Check that a phase does not have both startWith and any other start* set
+    *
+    * @param phase phase to check
+    */
+   private void checkStartWith(Phase phase) {
+      if (phase.startWithDelay != null && (!phase.startAfter.isEmpty() || !phase.startAfterStrict.isEmpty() || phase.startTime > 0)) {
+         throw new BenchmarkDefinitionException("Phase " + phase.name + " has both startWith and one of startAfter, startAfterStrict and startTime set.");
       }
    }
 
--- a/api/src/main/java/io/hyperfoil/api/config/Phase.java
+++ b/api/src/main/java/io/hyperfoil/api/config/Phase.java
@@ -3,11 +3,13 @@ package io.hyperfoil.api.config;
 import java.io.Serializable;
 import java.util.Collection;
 import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import io.hyperfoil.function.SerializableSupplier;
-
-import org.apache.logging.log4j.Logger;
-import org.apache.logging.log4j.LogManager;
 
 public final class Phase implements Serializable {
    protected static final Logger log = LogManager.getLogger(Phase.class);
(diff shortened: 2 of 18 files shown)


api/src/main/java/io/hyperfoil/api/config/PhaseReferenceDelay.java
```diff
@@ -0,0 +1,10 @@
+package io.hyperfoil.api.config;
+
+public class PhaseReferenceDelay extends PhaseReference {
```
lampajr, line 3: I don't feel very satisfied about this `PhaseReferenceDelay` class name, but I had to differentiate it with respect to `StartWithDelay` object (the one actually used in the Phase). I was thinking about adding optional `delay` field in the `PhaseReference` in order to avoid some duplication.. but I did not want to mix things up, wdyt?

api/src/main/java/io/hyperfoil/api/config/Phase.java
```diff
@@ -32,11 +32,9 @@ public final class Phase implements Serializable {
    public final Model model;
    public final boolean isWarmup;
    public final Map<String, SLA[]> customSlas;
+   public final StartWithDelay startWithDelay;
```
lampajr, line 35: After some reasoning I decided to implement this as a dedicated class, as it basically includes two different things:
- a phase ref as a string
- a delay
I decided this because in this way the phase ref will have eactly the same format used for all others `startAfter*` properties, so I could reused some of the methods to checks cyclic deps and so on.

If we don't want to have a complex object here, we could:
- split this in two separated fields, don't like as these two fields are correlated.
- create a single string as a result, similarly to what has been done for phase refs, e.g., `<phase-ref>:<delay>`

api/src/main/java/io/hyperfoil/api/config/BenchmarkBuilder.java
```diff
@@ -180,7 +184,7 @@ private void checkDeadlock(Phase phase, Map<String, Phase> phases) {
       toProcess.add(phase);
       while (!toProcess.isEmpty()) {
          Phase p = toProcess.poll();
-         Stream.concat(p.startAfter.stream(), p.startAfterStrict.stream()).forEach(name -> {
+         Stream.concat(Stream.concat(p.startAfter.stream(), p.startAfterStrict.stream()), p.startWithDelay != null ? Stream.of(p.startWithDelay.phase) : Stream.of()).forEach(name -> {
```
franz1981, line 187: please add a comment here as well.
You can check if it is worthy to make the deps list as part of the `Phase` API

test-suite/src/test/java/io/hyperfoil/benchmark/standalone/StartWithDelayTest.java
```diff
@@ -0,0 +1,130 @@
...
+import io.hyperfoil.http.api.HttpMethod;
+import io.hyperfoil.http.config.HttpPluginBuilder;
+import io.vertx.core.Handler;
+import io.vertx.core.http.HttpServerRequest;
+import io.vertx.ext.unit.junit.VertxUnitRunner;
+
+@Category(io.hyperfoil.test.Benchmark.class)
+@RunWith(VertxUnitRunner.class)
+public class StartWithDelayTest extends BaseBenchmarkTest {
+   protected final Logger log = LogManager.getLogger(getClass());
```
franz1981, line 38: static

test-suite/src/test/java/io/hyperfoil/test/YamlParserTest.java
```diff
@@ -258,27 +278,28 @@ private <T> void assertCondition(T object, Predicate<T> predicate) {
       assertThat(object).has(new Condition<>(predicate, ""));
    }
 
-   private Benchmark buildBenchmark(String s) {
-      ClassLoader classLoader = this.getClass().getClassLoader();
-      return buildBenchmark(classLoader.getResourceAsStream(s));
-   }
+//   private Benchmark loadScenario(String s) {
```
franz1981, line 281: delete

franz1981 (approved):
LGTM, thanks for this great contribution :)
One quick note; check if checkstyle isn't messing up with existing imports and feel free to add more tests if you want

api/src/main/java/io/hyperfoil/api/config/BenchmarkBuilder.java
lampajr: good idea, I will add `getDependencies()` method in `Phase` class to return all phases' dependencies.

lampajr:
> LGTM, thanks for this great contribution :) One quick note; check if checkstyle isn't messing up with existing imports and feel free to add more tests if you want

Thanks @franz1981, I applied your suggestions and I reverted some changes due to the formatter as not needed at all

lampajr:
> @lampajr Please, squash the commits

@jesperpedersen done!

jesperpedersen:
@franz1981 Let me know when you are happy with this

core/src/test/java/io/hyperfoil/core/session/BaseBenchmarkParserTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+public abstract class BaseBenchmarkParserTest {
+
+   protected Benchmark loadScenario(String name) {
+      try {
+         InputStream config = getClass().getClassLoader().getResourceAsStream(name);
+         Benchmark benchmark = loadBenchmark(config);
+         // Serialization here is solely for the purpose of asserting serializability for all the components
+         byte[] bytes = Util.serialize(benchmark);
+         assertThat(bytes).isNotNull();
```
franz1981, line 23: you can rehydrate it back and see if they are equivalent/can work

franz1981 (changes requested):

core/src/test/java/io/hyperfoil/core/session/BaseBenchmarkParserTest.java
lampajr: Hey @franz1981, you mean to serialize the Benchmark and then deserialize it checking if it will match the original Benchmark?

franz1981: Yep, in case in the future we would make some of the field `transient` (or deps field); probably is just overworrying on my side and we didn't have done it previously just feel free to ignore my advice :D

lampajr: Makes sense, in this case Ithink we could have two options:
- Implement `equals` in order to define our concept of bencharmk equality and just make the comparison everytime we load a benchmark
- Run all test that loads benchmark two times (one with the original benchamrk and one with the deserialized one)

Any preference?

franz1981: The second one increase by far the tests duration and still won't report what was "wrong" , so probably the former is more fine grain

franz1981: good point, let's keep them separated!

franz1981 (approved):
LGTM great work @lampajr !!!

core/src/test/java/io/hyperfoil/core/session/BaseBenchmarkParserTest.java
lampajr: Created issue https://github.com/Hyperfoil/Hyperfoil/issues/304 to keep track of this

jesperpedersen:
Merged.

PR #327, opened 2024-02-16
Update Infinispan to 13.0.21

wburns:
* Infinispan should be higher, but this is supported by vertx

Upgrade vert.x to latest which has a dep on Infinispan 13.0.17

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -92,7 +92,7 @@
         <version.slf4j>2.0.6</version.slf4j>
         <version.snakeyaml>2.0</version.snakeyaml>
         <version.vertx>4.5.4</version.vertx>
-        <version.infinispan>13.0.8.Final</version.infinispan>
+        <version.infinispan>13.0.21.Final</version.infinispan>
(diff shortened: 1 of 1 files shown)


wburns:
Changing back to draft as it seems something is broken

wburns:
Seems the vert.x and netty upgrade has broken something with some of the HTTP tests that I am familiar with atm.

franz1981:
It seems something between 4.4.8  (which work fine) and 4.5.1 vertx breaks it, I'm investigating

wburns:
Updated core of ISPN to 13.0.17 to match vertx.clustering. Note that without this there is a mismatch where some of the ISPN libraries used by Hyperfoil are 13.0.8 and others are 13.0.17, so this aligns them all.

wburns:
Forcing it to use ISPN 13.0.21 seems to pass all the tests as well, so no reason we should stop at 13.0.17.

wburns:
Yep, seems to all pass.

PR #328, opened 2024-02-27
Update Vert.x to 4.5.3 and Infinispan to 13.0.17

### LINKED ISSUE #319: Moving to latest Vertx 4.5
This can be a blocker for #281 due to https://github.com/eclipse-vertx/vert.x/pull/4933 which can be backported to latest 4.5 and get control of the gRPC connectivity, but we need to catch up first with latest vertx version.

franz1981:
This should supersede #327  and fixes #319 

@wburns I've addressed the most of the existing problems due to upgrading to a later vertx version, but need to check the whole CI to make sure it doesn't break other stuff.


franz1981:
I got few failures, which IDK yet if are related to the infinispan upgrade or vertx one, so I'm opening a separate PR to split the work in 2 ie #329

wburns:
Building this went fine with tests and I was able to run a benchmark via https://github.com/infinispan/infinispan-ansible-benchmark just fine

PR #329, opened 2024-02-27
Moving to latest Vertx 4.5.3 (Fixes #319)

### LINKED ISSUE #319: Moving to latest Vertx 4.5

franz1981:


franz1981:
@wburns any reason why you didn't upgrade to the very latest vertx?

franz1981:
Ok, so the failures we has in the infinispan upgrade was because of some changes in behaviour on infinispan, likely.

wburns:
> @wburns any reason why you didn't upgrade to the very latest vertx?

It was the latest when I created it :) We should definitely use 4.5.4 though as it just came out.

franz1981:
Given that this one is green if @barreiro is fine with it I would merge it

wburns:
Lets 👍 Also the PR title needs to change to 4.5.4 btw :)

franz1981:
This is ready to go than!

barreiro (approved):

PR #331, opened 2024-03-05
Update to JDK 17

wburns:
* Also updates all maven plugin usages

### DIFF
--- a/distribution/src/main/docker/Dockerfile
+++ b/distribution/src/main/docker/Dockerfile
@@ -1,16 +1,16 @@
-FROM registry.access.redhat.com/ubi8/openjdk-11
+FROM registry.access.redhat.com/ubi9/openjdk-17
 USER root
 RUN microdnf install -y less vim && curl -s -L https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 -o /usr/bin/jq && chmod a+x /usr/bin/jq
 EXPOSE 8090
 ENV EDITOR=vim HYPERFOIL_CONTAINER=true
 COPY maven/bin /deployment/bin/
 COPY maven/templates /deployment/templates
-COPY maven/examples /home/jboss/examples/
+COPY maven/examples /home/default/examples/
 ADD maven/hyperfoil-*-deps.tar /deployment/
 ADD maven/hyperfoil-*-hyperfoil.tar /deployment/
 RUN mkdir /deployment/extensions
-RUN chown -R jboss /deployment
-USER jboss
+RUN chown -R default /deployment
+USER default
 # executable permissions in bin/*.sh are lost somewhere
 RUN chmod a+x /deployment/bin/*.sh
 ENTRYPOINT [ "/deployment/bin/entrypoint.sh" ]
--- a/hotrod/pom.xml
+++ b/hotrod/pom.xml
@@ -192,7 +192,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
-                <version>3.2.4</version>
+                <version>3.5.2</version>
                 <executions>
                     <execution>
                         <phase>package</phase>
@@ -241,7 +241,7 @@
                         <configuration>
                             <target>
                                 <!-- We will just take the JDK11-specific classes and move them to regular locations
-                                     as Hyperfoil is JDK11+ only anyway. We won't need to mess with multi-release JAR then -->
+                                     as Hyperfoil is JDK17+ only anyway. We won't need to mess with multi-release JAR then -->
                                 <unzip src="${project.build.directory}/hyperfoil-hotrod-shaded.jar" dest="${project.build.directory}/shaded" />
                                 <move todir="${project.build.directory}/shaded/shaded">
                                     <fileset dir="${project.build.directory}/shaded/META-INF/versions/11/" />
@@ -274,7 +274,6 @@
                         <configuration>
                             <skip>${module.skipCopyDependencies}</skip>
                             <includeScope>runtime</includeScope>
-                            <excludeScope>test</excludeScope>
                             <excludeGroupIds>com.github.ben-manes.caffeine,org.infinispan,org.infinispan.protostream,org.wildfly</excludeGroupIds>
                             <silent>true</silent>
                         </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -73,8 +73,8 @@
     <properties>
 
         <!-- maven-compiler-plugin -->
-        <maven.compiler.target>11</maven.compiler.target>
-        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>17</maven.compiler.target>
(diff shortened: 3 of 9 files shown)


wburns:
> @wburns there are more places that need update. `.dockerfiles` in `distribution` folder and qyay.io action come to mind.

@barreiro I added both of those and found a couple more in README and parse-opts. Also I tried to run the Dockerfile but there is no `maven` directory so it would fail for me. Is there a required environment when running it?

hotrod/pom.xml
```diff
@@ -240,8 +240,8 @@
                         <phase>package</phase>
                         <configuration>
                             <target>
-                                <!-- We will just take the JDK11-specific classes and move them to regular locations
-                                     as Hyperfoil is JDK11+ only anyway. We won't need to mess with multi-release JAR then -->
+                                <!-- We will just take the JDK17-specific classes and move them to regular locations
+                                     as Hyperfoil is JDK17+ only anyway. We won't need to mess with multi-release JAR then -->
                                 <unzip src="${project.build.directory}/hyperfoil-hotrod-shaded.jar" dest="${project.build.directory}/shaded" />
                                 <move todir="${project.build.directory}/shaded/shaded">
                                     <fileset dir="${project.build.directory}/shaded/META-INF/versions/11/" />
```
franz1981, line 247: what's going on here? I see that things are shared into `11`?

pom.xml
```diff
@@ -571,38 +571,38 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-enforcer-plugin</artifactId>
-                    <version>3.0.0-M2</version>
+                    <version>3.4.1</version>
```
franz1981, line 574: @wburns I'm not familiar, but would be fine to add a dependabot from now on for such updates?

hotrod/pom.xml
wburns: Oh whoops meant to put 17. TBH, this code might be able to be deleted as I don't see a versions/11 or versions/17 directory after a build is ran.

pom.xml
wburns: 💯

barreiro:
LGTM

.github/workflows/quay.yml
```diff
@@ -13,11 +13,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
```
shivam-sharma7, line 15:
```suggestion
    - uses: actions/checkout@v4
```
use latest

distribution/src/main/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/openjdk-11
+FROM registry.access.redhat.com/ubi8/openjdk-17
```
shivam-sharma7, line 1:
```suggestion
FROM registry.access.redhat.com/ubi9/openjdk-17
```
use ubi9

shivam-sharma7 (changes requested):

wburns:
Sorry, was on leave for a while :)

Rebased and updated to use ubi9 for Docker.

distribution/src/main/docker/Dockerfile
lampajr, line 1: With the following diff on this file the image build seems working:
```bash
-COPY maven/examples /home/jboss/examples/
+COPY maven/examples /home/default/examples/
 ADD maven/hyperfoil-*-deps.tar /deployment/
 ADD maven/hyperfoil-*-hyperfoil.tar /deployment/
 RUN mkdir /deployment/extensions
-RUN chown -R jboss /deployment
-USER jboss
+RUN chown -R default /deployment
+USER default

```

lampajr (changes requested):
Hi @wburns,

I was checking the failure here and I think we need some other changes in the distribution Dockerfile, in particular we need to change `jboss` user to `default` as this seems the new default for `registry.access.redhat.com/ubi9/openjdk-17`.

distribution/src/main/docker/Dockerfile
wburns: Sorry, I didn't get a chance to check the errors yet, so thanks for the help!

I am swamped atm, but I will push the update in a little bit.

shivam-sharma7: @lampajr `jboss` to `default` ?

lampajr: @wburns No worries and no rush at all, hope this will save you some time :crossed_fingers: 

@shivam-sharma7 yep, they changed the default user in the `ubi9` image, which is actually `default`

wburns: Updated, thanks again 🙇

shivam-sharma7 (approved):
LGTM ✅

hotrod/pom.xml
lampajr: Hey @franz1981 @wburns  IIUC we should remove the `move` entry here because with this pull request, Hyperfoil will become JDK17+ only. Thus, moving everything from `${project.build.directory}/shaded/META-INF/versions/11/` to `${project.build.directory}/shaded/shaded` will override `jdk17` classes with `jdk11`-specific ones which is not what we want. wdyt?

pom.xml
lampajr: created https://github.com/Hyperfoil/Hyperfoil/issues/348 to keep track of this

lampajr:
@wburns @franz1981 I would roll out the `0.26.0` release before merging this one, but before that I would like to address this comment https://github.com/Hyperfoil/Hyperfoil/pull/331#discussion_r1606413193 (or simply clarify if that is not an issue)

hotrod/pom.xml
wburns: It isn't an issue and I actually completely remove it in https://github.com/Hyperfoil/Hyperfoil/pull/349

lampajr (approved):
> It isn't an issue and I actually completely remove it in https://github.com/Hyperfoil/Hyperfoil/pull/349

Ok, then lgtm!

lampajr:
Hi @wburns , sorry for the noise but with the latest release I had to slightly change one of the GitHub workflow (i.e., `quay.yml`), do you mind to rebase your branch on top of master and re-apply the same change on the new [.github/workflows/build-and-push-image.yml](https://github.com/Hyperfoil/Hyperfoil/blob/643568bc7731d381ea71d95ec929bc39046e2d3c/.github/workflows/build-and-push-image.yml)?

wburns:
Git was able to apply it on the rebase just fine with the move. Updated @lampajr !

lampajr (approved):
> Git was able to apply it on the rebase just fine with the move. Updated @lampajr !

Awesome!

I think we are ready to go with this :rocket:

franz1981 (approved):
Well done and thanks! 
For JDK 21 (which enable ZGC generational!) we need to update some plugins

lampajr:
Thanks a lot @wburns :rocket: 

>For JDK 21 (which enable ZGC generational!) we need to update some plugins

@franz1981 I will open another issue for this upgrade

Issue #333, opened 2024-03-28
Running Hyperfoil HotRod with remote agents sometimes gets stuck in READY state


wburns:
When running Hyperfoil controller and 2 agents on different machines many times it never starts despite running the first phase.

![image](https://github.com/Hyperfoil/Hyperfoil/assets/648175/49233224-bf25-48c4-9903-1929c2f32ea3)
 
Here is an image of when it happens. I am forced to kill the run and retry again. It happens around half the time.

The image is running the benchmark and it showing READY but never showing the phase durations left and then switching to stats seeing it is actually running and finally back to status which is still stuck.

franz1981:
The same benchmark (without controller) works correctly locally?
can you share the benchmark definition?

wburns:
It is not the same as #323 as I have seen that once, which kills the run completely.

I can very easily reproduce this with my EC2 benchmark, over half of the time recently :(

The benchmark definition is https://github.com/infinispan/infinispan-ansible-benchmark/blob/main/benchmarks/hotrod-benchmark.yaml.j2. You can ignore the agents part at the top, that is just so ansible can populate it for me automatically.

franz1981:
> over half of the time recently :

which seems to suggest to a timing issue: the benchmark definition seems one of the quickstarts, meaning it should't be problematic.

Do you have any logs from agent/controllers? To understand if something suspicious is happening?
Same for heap dump (stack trace won't be probably very useful there...), so I can inspect in which states agent/controller(s) believe to be.

> I have not tried with an agent, if that is what you meant by locally.

Yep, and (IDK if is possible), with a local setup too (single agent), to understand if can be reproduced locally, too.

wburns:
Yeah, I was going to try to dig into it further today when I get some time. I can get the logs of both and attach as well.

wburns:
[hfc-ip-172-31-12-151.log](https://github.com/Hyperfoil/Hyperfoil/files/14828545/hfc-ip-172-31-12-151.log)

Attached is a log file. If you notice runs 0000-0002 all ran fine. But then 0003 didn't, the rampup phases never received a status update. So it sounds like the agents aren't sending the messages back in some cases.

Looking at the agent log it seems to just forever do the rampupGet and rampupPut phases and always prints "no progress, not finished." So it looks like a bug in the agent code somehow

[...]

PR #334, opened 2024-04-02
Provide openConnections method a blocking handler for plugins

wburns:
* HotRod new RemoteCacheManger would block event loop thread

Fixes https://github.com/Hyperfoil/Hyperfoil/issues/333

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/impl/LocalSimulationRunner.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/LocalSimulationRunner.java
@@ -15,6 +15,7 @@ import io.hyperfoil.api.config.BenchmarkDefinitionException;
 import io.hyperfoil.api.config.Phase;
 import io.hyperfoil.api.session.PhaseInstance;
 import io.hyperfoil.core.impl.statistics.StatisticsCollector;
+import io.vertx.core.Future;
 
 public class LocalSimulationRunner extends SimulationRunner {
    private final StatisticsCollector.StatisticsConsumer statsConsumer;
@@ -45,7 +46,15 @@ public class LocalSimulationRunner extends SimulationRunner {
 
       CountDownLatch latch = new CountDownLatch(1);
       init();
-      openConnections(result -> latch.countDown());
+      openConnections(callable -> {
+               try {
+                  callable.call();
+               } catch (Exception e) {
+                  return Future.failedFuture(e);
+               }
+               return Future.succeededFuture();
+            },
+            result -> latch.countDown());
       try {
          latch.await();
          statsExecutor.scheduleAtFixedRate(this::collectStats, 0, benchmark.statisticsCollectionPeriod(), TimeUnit.MILLISECONDS);
--- a/clustering/src/main/java/io/hyperfoil/clustering/AgentVerticle.java
+++ b/clustering/src/main/java/io/hyperfoil/clustering/AgentVerticle.java
@@ -4,6 +4,9 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
 import io.hyperfoil.Hyperfoil;
 import io.hyperfoil.api.config.Benchmark;
 import io.hyperfoil.api.session.PhaseInstance;
@@ -11,10 +14,10 @@ import io.hyperfoil.clustering.messages.AgentControlMessage;
 import io.hyperfoil.clustering.messages.AgentHello;
 import io.hyperfoil.clustering.messages.AgentReadyMessage;
 import io.hyperfoil.clustering.messages.ErrorMessage;
-import io.hyperfoil.core.util.CountDown;
-import io.hyperfoil.core.impl.SimulationRunner;
 import io.hyperfoil.clustering.messages.PhaseChangeMessage;
 import io.hyperfoil.clustering.messages.PhaseControlMessage;
+import io.hyperfoil.core.impl.SimulationRunner;
+import io.hyperfoil.core.util.CountDown;
 import io.hyperfoil.impl.Util;
 import io.hyperfoil.internal.Properties;
 import io.vertx.core.AbstractVerticle;
@@ -25,9 +28,6 @@ import io.vertx.core.eventbus.MessageConsumer;
 import io.vertx.core.eventbus.ReplyException;
 import io.vertx.core.eventbus.ReplyFailure;
 import io.vertx.core.impl.VertxInternal;
-
-import org.apache.logging.log4j.Logger;
-import org.apache.logging.log4j.LogManager;
 
 public class AgentVerticle extends AbstractVerticle {
    private static Logger log = LogManager.getLogger(AgentVerticle.class);
@@ -262,7 +262,7 @@ public class AgentVerticle extends AbstractVerticle {
          connectionStatsSender.send();
       });
 
-      runner.openConnections(result -> {
+      runner.openConnections(vertx::executeBlocking, result -> {
          if (result.succeeded()) {
             eb.send(Feeds.RESPONSE, new AgentReadyMessage(deploymentID(), runId));
          } else {
--- a/core/src/main/java/io/hyperfoil/core/api/PluginRunData.java
+++ b/core/src/main/java/io/hyperfoil/core/api/PluginRunData.java
@@ -1,7 +1,9 @@
 package io.hyperfoil.core.api;
 
 import java.time.Clock;
+import java.util.concurrent.Callable;
 import java.util.function.Consumer;
+import java.util.function.Function;
 
 import io.hyperfoil.api.config.Scenario;
 import io.hyperfoil.api.session.Session;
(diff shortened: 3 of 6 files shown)


core/src/main/java/io/hyperfoil/core/impl/LocalSimulationRunner.java
```diff
@@ -45,7 +46,15 @@ public void run() {
 
       CountDownLatch latch = new CountDownLatch(1);
       init();
-      openConnections(result -> latch.countDown());
+      openConnections(callable -> {
+               try {
+                  callable.call();
```
wburns, line 51: I _think_ this okay as there is no event loop around with a local simulation run, but @franz1981 can you confirm?

franz1981: Will tell you tomorrow, but ATM I would say no

franz1981: Let me check how I could configure block hound for our tests

franz1981:
I'm checking what's going on and found that:
1. https://github.com/Hyperfoil/Hyperfoil/blob/77e538713b169fcad0d7773ec3d64b098f68e260/clustering/src/main/java/io/hyperfoil/clustering/AgentVerticle.java#L265 is assuming that `assert context.isEventLoopContext()` which doesn't mean is running from the even loop (according to the doc of Vertx), but given that it is issued as consumer of the Vertx `EventBus` it should be running from the event loop
2. when `openConnections` is issued, it will call into `HotRodRunData::openConnections`, than https://github.com/Hyperfoil/Hyperfoil/blob/3d5d7751df384977d5e51e1709f83790e861495f/hotrod/src/main/java/io/hyperfoil/hotrod/connection/HotRodRemoteCachePoolImpl.java#L49 which, on `RemoteCacheManager`'s `start` can hit this:
```java
  java.lang.Thread.State: RUNNABLE
	  at org.infinispan.client.hotrod.impl.transport.netty.ChannelFactory.pingServersIgnoreException(ChannelFactory.java:197)
	  at org.infinispan.client.hotrod.impl.transport.netty.ChannelFactory.start(ChannelFactory.java:150)
	  at org.infinispan.client.hotrod.RemoteCacheManager.actualStart(RemoteCacheManager.java:362)
	  at org.infinispan.client.hotrod.RemoteCacheManager.start(RemoteCacheManager.java:316)
	  at org.infinispan.client.hotrod.RemoteCacheManager.<init>(RemoteCacheManager.java:174)
	  at org.infinispan.client.hotrod.RemoteCacheManager.<init>(RemoteCacheManager.java:131)
	  at io.hyperfoil.hotrod.connection.HotRodRemoteCachePoolImpl.start(HotRodRemoteCachePoolImpl.java:49)
	  at io.hyperfoil.hotrod.HotRodRunData.openConnections(HotRodRunData.java:55)
	  at io.hyperfoil.core.impl.SimulationRunner.openConnections(SimulationRunner.java:163```
which await `PingOperation`s to reach out each server, while running it from the Netty event loop,see
```java
private void pingServersIgnoreException() {
      Collection<InetSocketAddress> servers = topologyInfo.getAllServers();
      for (SocketAddress addr : servers) {
         // Go through all statically configured nodes and force a
         // connection to be established and a ping message to be sent.
         try {
            await(fetchChannelAndInvoke(addr, operationsFactory.newPingOperation(true)));
         } catch (Exception e) {
            // Ping's objective is to retrieve a potentially newer
            // version of the Hot Rod cluster topology, so ignore
            // exceptions from nodes that might not be up any more.
            if (log.isTraceEnabled())
               log.tracef(e, "Ignoring exception pinging configured servers %s to establish a connection",
                     servers);
         }
      }
   }
   ```
which `PingOperation` is going to be executed as soon as the connection to servers are established, but running from the event loop thread configured:
```java
"epollEventLoopGroup-5-1@9906" tid=0x35 nid=NA runnable
  java.lang.Thread.State: RUNNABLE
	  at org.infinispan.client.hotrod.impl.operations.PingOperation.invoke(PingOperation.java:53)```
```
In short, my perception is that `openConnections` doesn't assume to be non-blocking, so I'm proposing a different fix here,ie https://github.com/franz1981/Hyperfoil/tree/never_blocking_el

IDK if is better or not, but I see that the local runnner doesn't assume opeening the connections to be a non-blocking operation, so I see as a responsability of the caller (which knows if is running on an event loop - which should never be blocked) to handle such.
Let me know if this is still fixing your case; I'm trying to write a test for this, regardless (both fixes should work with the test)

wburns:
@franz1981 while I am certain your commit will fix the issue, I personally don't like the approach. The openConnections Plugin API already implies it should be non blocking with the method declaration as shown here: 
https://github.com/Hyperfoil/Hyperfoil/blob/ac3ad2eaa929d363d8f7e850f6c1d4d40b594ab4/core/src/main/java/io/hyperfoil/core/api/PluginRunData.java#L15

The argument to me says the implementor shouldn't block and pass any operations that could block to the consumer so it can be notified after they are all done.

franz1981:
Yeah, but at this point, if we assume that it is the duty of the implementor to not block (because its API suggest that it shouldn't be blocking), you shouldn't allow to pass vertx::executeBlocking to it, because it would, again, give the responsibility to the caller, despite the async API.

If we want openConnections to be non blocking, it should be the spwcofoc HotRod implementation thst should take care to do it (being non blocking as expected) without changing its API.

In addition to what said, I would add some comment, on the API, the caller, or the implementor that explain what's going and why.

wburns:
> Yeah, but at this point, if we assume that it is the duty of the implementor to not block (because its API suggest that it shouldn't be blocking - I have no strong opinion on it TBH, and probably only @rvansa knows it), you shouldn't allow to pass vertx::executeBlocking to it, because it would, again, give the responsibility to the caller, despite the expected async API (& behavior).

> If we want openConnections to be non blocking, it should be the HotRod implementation that should take care to do it (being non blocking as expected) without changing its API (or via some other base class which provides the mechanism to do it right).

At this point implementors have to be non blocking to work properly. I guess we could relax the requirements for lifecycle operations such as start and stop.

The problem about forcing the implementation to handle the blocking is, we do not provide the tools to them. Namely there is no access to a blocking executor or way to run blocking code. This in turn means the implementor has to create their own threads, which seems very unfriendly. The nicest way to me was to provide a facility to run some code in a blocking thread (aka. vertx::executeBlocking). I chose that abstraction to not expose the plugin to vertx, but I am open to other avenues if you want.

> Clearly if we assume openConnections to be async, we should (if broken) fix the close side as well, to make sure the state machine is right (not in this PR, unless making it non blocking mess up with its state).

Yeah, I had debated about changing close as well, as that would break the non blocking contract as well. However, it doesn't actively cause the test to not run properly, so I left it for another day.

> In addition to what said, I would add some comment, on the API, the caller, or the implementor that explain what's going on and why.

Definitely agree, I didn't have a chance to add the comment yesterday as I had to leave immediately after I submitted. I can add this to the PR though, for sure.


TLDR; I guess I am fine with making start/stop run on a blocking thread and allowing the implementation to be blocking in those cases, it just seems inconsistent as everything else has to be non blocking in the plugin.

franz1981:
Yeah I tried to check if we pass somehow any blocking executor (from vertx) on the implementors but seems not :/ 
So I agree that we have to use a generic mechanism (or an executor-like parameter too, at this point is the same..);

I have noticed that:
- HotRodRunData::openConnections was ignoring the passed promiseCollector (meaning that if something break - now asynchronously, none knows...)
- HotRodRemoteCachePoolImpl::start is manipulating thread unsafe maps eg remoteCaches, which could be accessed by HotRodRequestStep::invoke at pool.getRemoteCache; if we are now asynx we risk thst the connection won't be established, the access will race AND if lucky (while unlucky) no remoteCache will be found likely throwing NPE on the step (which, remember, has started regardless the pool start completion!)

wburns:
> Yeah I tried to check if we pass somehow any blocking executor (from vertx) on the implementors but seems not :/ So I agree that we have to use a generic mechanism (or an executor-like parameter too, at this point is the same..);
> 
> I have noticed that:
> 
> * HotRodRunData::openConnections was ignoring the passed promiseCollector (meaning that if something break - now asynchronously, none knows...)

It should throw an exception at least, but yes I agree and is fixed in this PR.

> * HotRodRemoteCachePoolImpl::start is manipulating thread unsafe maps eg remoteCaches, which could be accessed by HotRodRequestStep::invoke at pool.getRemoteCache; if we are now async, we risk that the connection won't be established, the access will race AND if lucky (while unlucky) no remoteCache will be found likely throwing NPE on the step (which, remember, has started regardless the pool start completion!)

I am not certain, but I don't think Invoke can be ran concurrently with start as openConnections doesn't call handle until all the pools are created. However, the variable is indeed not thread safe so there is a small chance that the other thread may not see the updated value, although I doubt it wouldn't hit a memory barrier to sync that up someplace else. But we can easily change the map to a ConcurrentMap to be safe.

franz1981:
Yeah I tried to check if we pass somehow any blocking executor (from vertx) on the implementors but seems not :/ 
So I agree that we have to use a generic mechanism (or an executor-like parameter too, at this point is the same..);

I have noticed that:
- HotRodRunData::openConnections was ignoring the passed promiseCollector (meaning that if something break - now asynchronously, none knows...)
- HotRodRemoteCachePoolImpl::start is manipulating thread unsafe maps eg remoteCaches, which could be accessed by HotRodRequestStep::invoke at pool.getRemoteCache; if we are now asynx we risk thst the connection won't be established, the access will race AND if lucky (while unlucky) no remoteCache will be found likely throwing NPE on the step (which, remember, has started regardless the pool start completion!)

franz1981:
I can easily verify if NPE can happen actually, let me use your branch

rvansa:
>  if we assume that it is the duty of the implementor to not block (because its API suggest that it shouldn't be blocking - I have no strong opinion on it TBH, and probably only @rvansa knows it), you shouldn't pass any vertx::executeBlocking to it, because it would, again, give the responsibility to the caller, despite the expected async API (& behavior).

Yes, I assumed this is running from Vert.x event-loop and hence shouldn't block. I don't see any issue with passing a blocking executor as long as eventually the callback is invoked. As this is executed only during initialization, you won't break any benchmark by changing this - as long as you don't break the Vert.x wiring.

> The problem about forcing the implementation to handle the blocking is, we do not provide the tools to them. Namely there is no access to a blocking executor or way to run blocking code. This in turn means the implementor has to create their own threads, which seems very unfriendly.

Initially the API was written for HTTP, and async implementation in general, assuming that any reasonable network protocol API will be non-blocking. It's a defect on the target library side to let the thread wait :)

These callback-based APIs look a bit unwieldy anyway, though - change them at will if that suits you better :)

franz1981:
Many thanks @rvansa for it!

wburns:
Added the docs for the openConnections SPI method. Unfortunately, after tomorrow I will not be able to continue working on this.

franz1981:
No need, you can merge it as it is

franz1981:
Or just wait till I check what's left, if any (I cannot see what could be)

franz1981 (approved):
LGTM!

franz1981:
Thanks for this fix, @wburns and well done, it was tricky to find!

PR #336, opened 2024-04-19
Throw exception for misconfigured multi-host HTTP benchmark

jabolina:
Throw an exception for an HTTP benchmark with multi-hosts where the step doesn't define which host to run. It seems this check was the intended behavior.

However, I think it would be nice to make this "pinning" to a host optional. Otherwise, we would need to define the steps for each host. I'll open this first PR to make sure the exception is thrown and later look at how it is done on Hot Rod to handle the multi-hosts. And if it makes sense, we can update in a follow-up.


franz1981:
Many thanks!!

franz1981:
Running the test suite, but it looks ok to me

Issue #344, opened 2024-05-15
Make automatic reports opening configurable


lampajr:
At the moment if `Hyperfoil` is started locally and not in a container, when you create the report it will automatically open it in the browser:

https://github.com/Hyperfoil/Hyperfoil/blob/5dbd941d579c47c1f5a03021468f266934ba9e80/cli/src/main/java/io/hyperfoil/cli/commands/Report.java#L63-L65

There might be cases where users don't want to open the reports automatically, especially becase it will keep the Hyperfoil CLI stuck:

![image](https://github.com/Hyperfoil/Hyperfoil/assets/26715795/989e7d4d-462f-4c01-aed4-2666bf584338)

**Proposal**

I'd suggest to add an additional option to skip the automated report opening.
* name: `--silent`
* usage: `report 0000 --assumeYes --destination=/tmp --silent`

Even if I like `--silent`, the short alternative would collide with the `--source` option, therefore any other suggestion is appreciated like `--no-open` or any other.

franz1981:
it seems a very useful one! +1

PR #345, opened 2024-05-15
Disable automatic report opening with --silent option

lampajr:
Fixes https://github.com/Hyperfoil/Hyperfoil/issues/344

As first proposal I named the new option `--silent` without short version.

I tested it locally by starting `./distribution/target/distribution/bin/cli.sh` and running:
```bash
report 0000 --destination=/tmp --assumeYes --silent
```

here the result:
image

Without the `--silent` option the behavior is the default one:
image

I am not aware of any unit test for these kind CLI-related changes, if I missed something lmk :pray:


franz1981 (approved):
LGTM, thanks!

PR #350, opened 2024-05-29
Avoid byte[] allocation per event object

wburns:
This avoids copying a byte[] per event object send via the event bus to lower some allocation noise.

TBH if we convert RequestStatsMessage to use its own MessageCodec instead it would remove even more noise since it is sent every second. The ByteArrayOutputStream used per message wastes quite a bit of allocations resizing the internal byte[], but I figured this PR was small and simple enough to do as a first step.


franz1981 (approved):
LGTM!!

lampajr (approved):
lgtm as well!

@franz1981 do we want to include this in the next release?

lampajr:
Going to merge this as the `0.26` release has been already rolled out.

Thanks again @wburns :heart:

Issue #352, opened 2024-06-02
Fix container image build and push workflow


lampajr:
Fix the workflow that aims to build and push the Hyperfoil container image to quay.io: https://github.com/Hyperfoil/Hyperfoil/blob/master/.github/workflows/quay.yml 

 

Required changes:

- Trigger the workflow on every push on master branch --> simply push the container image as -SNAPSHOT version
- Trigger the workflow on every tag push --> build and push image as the tag version and as latest
- If the workflow is triggered by a tag, create a GitHub release and upload the hyperfoil-<VERSION>.zip.

PR #353, opened 2024-06-02
Fix container image build & push workflow

lampajr:
Fixes https://github.com/Hyperfoil/Hyperfoil/issues/352

**Proposed changes:**

- Trigger the workflow on `master` and _tag_ pushes (ignoring if the changes are only on text files)
- Update `docker/login-action` to version `v3`
- Use `softprops/action-gh-release@v2` instead of the archived `actions/create-release@v1`
  - With this new action we can upload the additional asset directly with that action


franz1981 (approved):

franz1981:
Thanks @lampajr !

Issue #342, opened 2024-05-13
OpenModelPhase cannot go past 1000 rps


franz1981:
Due to the millisecond delta precision, users cannot exceed 1000 rps for constantRate, rumpUp-kind of workload

franz1981:
See https://github.com/Hyperfoil/Hyperfoil/blob/5dbd941d579c47c1f5a03021468f266934ba9e80/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java#L350

franz1981:
It would be nice to make uses of a GC-free microseconds time from epoch timer, if possible, too, without using JNI/JNA.
The simpler approach is to adapt https://github.com/apache/activemq-artemis/blob/b9bb494b580ce59d41e63d4cd93ffdb07dc5dfaa/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/perf/MicrosClock.java#L23

Which relies on https://github.com/openjdk/jdk/blob/4785461f61d8f5c7444d2e6fd90f1e083dbc6fe4/src/hotspot/share/prims/jvm.cpp#L258 but I am not sure it is reliable across machines, if compared to other, more atomic ways.

franz1981:
this still means @wburns this is no longer a blocker for you, wdyt?

willr3:
I think I follow along the with the thought process. Are we seeing that it loops 1000 times at the beginning of each millisecond tick and instead we need it to spread the 1000 new requests over the full millisecond?

franz1981:
Exactly, beacuse right now, given that we can schedule the next fire time only at millisecond granularity, we just schedule the batch required to cover for the current missing ones (assuming we're keeping up with the avg expected rate)

i.e. 100K rps -> 100 rpms (rate per millisecond)

wburns:
This sounds like it could lead to spikey latencies if the user/second is a good chunk more than 1K/s. For example if your goal was 10K/s that means you will have 10 requests immediately at the beginning of a millisecond? In some of my tests my average latency is < 200 us so that would be make a big impact, where my client may be slammed for 200-400 us and then doing nothing for the remainder of that millisecond?

Also I assume this doesn't affect the number of agents. That is the 1K/s is only per agent, so technically you can do up to 1000 * N operations/s per agent and not see this behavior?

franz1981:
[...]

PR #357, opened 2024-06-05
Rate generators refactoring and compensation unification

franz1981:
This is required for https://github.com/Hyperfoil/Hyperfoil/issues/342 because there are no tests for not point poisson rate generators, nor unit tests to verify that the rate generators are working as expected.

I'll keep on working on this to add more deterministic tests.

The most relevant change here is the change in the compensation logic at [8c9fa8ef60841d2b60eedd988ab3ed57aeb1610c for the ](https://github.com/Hyperfoil/Hyperfoil/commit/8c9fa8ef60841d2b60eedd988ab3ed57aeb1610c#diff-088ca7616526932d6ec3865d81225f8f40704022d96dacfd1dd461426de4dc7fR34) which unify the fuzzy/precise generators, making the latter to not give up on compensating users when the pool is exhausted and the removal of https://github.com/Hyperfoil/Hyperfoil/commit/8c9fa8ef60841d2b60eedd988ab3ed57aeb1610c#diff-ebe33043584e91c6a83692dedff9dcba583b224468b420c05364e0259a8c1dc1L29 because (unless bugs!) `nextRequiredUsers` should be always bigger than `startedOrThrottledUsers` and the delta is supposed to be the amount of required users from the tick to the next scheduled one.

For this last statement I would ask @willr3 and @rvansa if they got historical memories why the `Math.max` was required.

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/impl/OpenModelPhase.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/OpenModelPhase.java
@@ -1,0 +1,116 @@
+package io.hyperfoil.core.impl;
+
+import io.hyperfoil.api.config.Model;
+import io.hyperfoil.api.config.Phase;
+import io.hyperfoil.api.session.Session;
+import io.hyperfoil.core.impl.rate.FireTimeListener;
+import io.hyperfoil.core.impl.rate.RateGenerator;
+import io.netty.util.concurrent.EventExecutorGroup;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * This is a base class for Open Model phases that need to compensate users based on the available ones in the session pool.
+ * <br>
+ * The notion of being "throttled" users requires some explanation:
+ * <ul>
+ *     <li>a user requires a {@link Session} to run, hence the two terms are used to denote the same concept</li>
+ *     <li>starting a {@link Session} doesn't mean immediate execution, but scheduling a deferred start in the {@link Session#executor()}</li>
+ *     <li>given that {@link Session}s are pooled, being throttled means that no available instances are found by {@link #onFireTimes(long)}</li>
+ *     <li>when a {@link Session} finishes, {@link #notifyFinished(Session)} can immediately restart it if there are
+ *     throttled users, preventing it to be pooled</li>
+ * </ul>
+ * <p>
+ * The last point is crucial because it means that a too small amount of pooled sessions would be "compensated" only
+ * when a session finished, instead of when {@link #sessionPool} has available sessions available.
+ */
+final class OpenModelPhase extends PhaseInstanceImpl implements FireTimeListener {
+
+    private final int maxSessions;
+    private final AtomicLong throttledUsers = new AtomicLong(0);
+    private final RateGenerator rateGenerator;
+    private final long relativeFirstFireTime;
+
+    OpenModelPhase(RateGenerator rateGenerator, Phase def, String runId, int agentId) {
+        super(def, runId, agentId);
+        this.rateGenerator = rateGenerator;
+        this.maxSessions = Math.max(1, def.benchmark().slice(((Model.OpenModel) def.model).maxSessions, agentId));
+        this.relativeFirstFireTime = rateGenerator.lastComputedFireTimeMs();
+    }
+
+    @Override
+    public void reserveSessions() {
+        if (log.isDebugEnabled()) {
+            log.debug("Phase {} reserving {} sessions", def.name, maxSessions);
+        }
+        sessionPool.reserve(maxSessions);
+    }
+
+    @Override
+    protected void proceedOnStarted(final EventExecutorGroup executorGroup) {
+        long elapsedMs = System.currentTimeMillis() - absoluteStartTime;
+        long remainingMsToFirstFireTime = relativeFirstFireTime - elapsedMs;
+        if (remainingMsToFirstFireTime > 0) {
+            executorGroup.schedule(() -> proceed(executorGroup), remainingMsToFirstFireTime, TimeUnit.MILLISECONDS);
+        } else {
+            // we are not enforcing to be called from an event loop thread here, and indeed tests uses the main thread:
+            proceed(executorGroup);
+        }
+    }
+
+    @Override
+    public void proceed(final EventExecutorGroup executorGroup) {
+        if (status.isFinished()) {
+            return;
+        }
+        long realFireTimeMs = System.currentTimeMillis();
+        long elapsedTimeMs = realFireTimeMs - absoluteStartTime;
+        // the time should flow forward: we can have some better check here for NTP and maybe rise a warning
+        assert elapsedTimeMs >= rateGenerator.lastComputedFireTimeMs();
+        long expectedNextFireTimeMs = rateGenerator.computeNextFireTime(elapsedTimeMs, this);
+        // we need to make sure that the scheduling decisions are made based on the current time
+        long rateGenerationDelayMs = System.currentTimeMillis() - realFireTimeMs;
+        assert rateGenerationDelayMs >= 0;
+        // given that both computeNextFireTime and onFireTimes can take some time, we need to adjust the fire time
+        long scheduledFireDelayMs = Math.max(0, (expectedNextFireTimeMs - elapsedTimeMs) - rateGenerationDelayMs);
+        if (trace) {
+            log.trace("{}: {} after start, {} started ({} throttled), next user in {} ms, scheduling decisions tooks {} ms", def.name, elapsedTimeMs,
+                    rateGenerator.fireTimes(), throttledUsers, scheduledFireDelayMs, rateGenerationDelayMs);
+        }
+        if (scheduledFireDelayMs <= 0) {
+            // we're so late that there's no point in bothering the executor with timers
+            executorGroup.execute(() -> proceed(executorGroup));
+        } else {
+            executorGroup.schedule(() -> proceed(executorGroup), scheduledFireDelayMs, TimeUnit.MILLISECONDS);
+        }
+    }
+
+    @Override
+    public void onFireTime() {
+        if (!startNewSession()) {
+            throttledUsers.incrementAndGet();
+        }
+    }
+
+    @Override
+    public void notifyFinished(Session session) {
+        if (session != null && !status.isFinished()) {
+            long throttled = throttledUsers.get();
+            while (throttled != 0) {
+                if (throttledUsers.compareAndSet(throttled, throttled - 1)) {
+                    // TODO: it would be nice to compensate response times
+                    // in these invocations for the fact that we're applying
+                    // SUT feedback, but that would be imprecise anyway.
+                    session.start(this);
+                    // this prevents the session to be pooled
+                    return;
+                } else {
+                    throttled = throttledUsers.get();
+                }
+            }
+        }
+        super.notifyFinished(session);
+    }
+
+}
--- a/core/src/main/java/io/hyperfoil/core/impl/rate/FireTimeListener.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/rate/FireTimeListener.java
@@ -1,0 +1,14 @@
+package io.hyperfoil.core.impl.rate;
+
+@FunctionalInterface
+public interface FireTimeListener {
+
+    void onFireTime();
+
+    default void onFireTimes(long count) {
+        for (long i = 0; i < count; i++) {
+            onFireTime();
+        }
+    }
+
+}
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -129,6 +129,12 @@
             <artifactId>assertj-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-math3</artifactId>
+            <version>3.6.1</version>
+            <scope>test</scope>
+        </dependency>
 
     </dependencies>
 
--- a/core/src/main/java/io/hyperfoil/core/impl/OpenModel.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/OpenModel.java
@@ -1,0 +1,31 @@
+package io.hyperfoil.core.impl;
+
+import io.hyperfoil.api.config.Model;
+import io.hyperfoil.api.config.Phase;
+import io.hyperfoil.api.session.PhaseInstance;
+import io.hyperfoil.core.impl.rate.RateGenerator;
+
+final class OpenModel {
+
+   public static PhaseInstance constantRate(Phase def, String runId, int agentId) {
+      var model = (Model.ConstantRate) def.model;
+      double usersPerSec = def.benchmark().slice(model.usersPerSec, agentId);
+      if (model.variance) {
+         return new OpenModelPhase(RateGenerator.poissonConstantRate(usersPerSec), def, runId, agentId);
+      } else {
+         return new OpenModelPhase(RateGenerator.constantRate(usersPerSec), def, runId, agentId);
+      }
+   }
+
+   public static PhaseInstance rampRate(Phase def, String runId, int agentId) {
+      var model = (Model.RampRate) def.model;
+      double initialUsersPerSec = def.benchmark().slice(model.initialUsersPerSec, agentId);
+      double targetUsersPerSec = def.benchmark().slice(model.targetUsersPerSec, agentId);
+      long durationMs = def.duration;
+      if (model.variance) {
+         return new OpenModelPhase(RateGenerator.poissonRampRate(initialUsersPerSec, targetUsersPerSec, durationMs), def, runId, agentId);
+      } else {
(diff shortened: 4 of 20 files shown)


franz1981:
This PR is still in draft to get feedbacks on the naming (which are always a PITA) and if the "new" hierarchy better fit.The nice thing of reorganizing the code has delivered, is that I've reduced the state space to the bare minimum and I could apply aggressive pre-computations (i can add more, actually) while solving the equations of the rate generators.

franz1981:
While reading the whole changes I see I have an idea to further unify the rate generators. I will work on this this afternoon, but the rest of the changes mentioned in the pr descriptions still hold

core/src/main/java/io/hyperfoil/core/impl/user/PointPoissonRampUserGenerator.java
```diff
@@ -0,0 +1,89 @@
...
+        this.scheduledUser.fireTime = nextFireTimeMs(0);
+    }
+
+    @Override
+    protected long nextFireTimeMs(final long elapsedTimeMs) {
+        // we're solving quadratic equation coming from t = (duration * -log(rand))/(((t + now) * (target - initial)) + initial * duration)
+        if (tinyACoef) {
+            // prevent division 0f/0f
+            return PointPoissonConstantUserGenerator.fireTimeMs(random, initialUsersPerSec, elapsedTimeMs);
+        }
```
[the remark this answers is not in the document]
franz1981: Happy of this comment because indipendently we arrive to the same conclusion! Many thanks!

franz1981:
@lampajr 
>  I am not 100% sure if that can have some side effects, e.g., accentuate the gap (the gap itself is fine) between two phases as in the below screenshots (maybe it might be imperceptible to the user):

I have taken a look at ElasticPool::acquire amd indeed it could be costly enough that if the phase lag behind of "enough" users this can make it further distant, good point!

I see different ways to solve it:
1. Always assume startNewSessions to be costly and immediately return back to generate new users (via the rate generators) in case the next fire time is already passed: it is partially dangerous because we risk to monopolize the current Thread making the new sessions to never start, in case there is a single event loop. 
2. Always assume startNewSessions to be costly but just re-evaluate if the delta computed is still valid and if not (meaning it should be immediately scheduled or just scheduled earlier) and change the scheduling delta accordingly, leaving the next scheduled fire to injecting the lagged behind users 
3. Change the API of the generators to accept a UserControl interface which allows the rate generator  to immediately schedule the new Session while looping

Regardless, I think we should improve the ElasticPool to:
1. Use per event loop queues to make session's start and acquisition less costly (and contended)
2. Use a better queue under contention (which is not obvious to find - let's see)
3. Change the pool API to be more batchy: int acquire(Consumer<? super T> acquired, int limit). This will eventually improve the behaviour under contention, regardless 

What I see is that we should have a jmh module to verify how the core classes behave and how much they cost, too.

Last but not least: how did you produces the chart? Which rate and what you have observed?

franz1981:
I will add more tests as the previous comment will land to a more stable API or risk to keep changing them over and over

franz1981:
FYI @lampajr I've added 9e8c5fc2b41f322ee1e45bff505d3089baee36d8 to address the https://github.com/Hyperfoil/Hyperfoil/pull/357#issuecomment-2156186667 as per the second point 
i.e.
> 2. Always assume startNewSessions to be costly but just re-evaluate if the delta computed is still valid

This should help to address something I've mentioned at:
> In the current master code the elapsed time wasn't re-evaluating the current/real elapsed time while injecting users

Indeed, even in the original code in `master`, the loop to compute the next fire time just needed to past over the existing snapshot of elapsed time, which is maybe no longer valid after we `Session`'s acquisitions (being a costy operation, regardless!).
Hopefully, thanks to this change, what you have observed should be mitigated (fingers crossed), but I would like some test which we can put in to for this, given that it can impact users and we need to keep it under control, for regression purposes.

core/src/main/java/io/hyperfoil/core/impl/OpenModelPhase.java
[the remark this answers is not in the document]
franz1981: Yep, I think that should be the case as well; just to avoid having a single giant god class, untestable, as it is now

franz1981:
I've added 0f8b33b290e2e4de98150727776f32e76fabbbb2 on top of this branch (in a separate new branch, at https://github.com/franz1981/Hyperfoil/tree/rate_generators_session_pool) to improve the session's acquisition as  per

https://github.com/Hyperfoil/Hyperfoil/pull/357#issuecomment-2156186667 i.e.:

> 2. Use a better queue under contention (which is not obvious to find - let's see)

Given that having per-event loop queues would requires some specific knowledge of the different event loops, upfront: a thing which could be done, TBH, but would make `ElasticPool` a bit less generic - which seems ok given that only `Session`s are used there....

franz1981:
I've just noticed something odd: see https://github.com/Hyperfoil/Hyperfoil/blob/master/api/src/main/java/io/hyperfoil/api/connection/Request.java#L63 and https://github.com/Hyperfoil/Hyperfoil/blob/master/http/src/main/java/io/hyperfoil/http/steps/SendHttpRequestStep.java#L65: TLDR it seems to me that requests have no memory of the so called "intended" start time, inherited by the sessions run on the phase instance, which makes us prone to coordinated omission (!!) on master too.
This probably worth its own issue, but the current solution, which can spin while computing the next fire time, just further increase the skew between the load generator fire time (which is the intended one we should report) and the actual one collected on session's start, further exposing us to this problem. 
I'll think about it but it seems to me that:
- we should measure if the skew is relevant (it seems by the test from @lampajr )
- we should atore the phase instance expected fire time into the Session, to allow who run within it (e.g. Request) to use it in the statistics, accounting for the infrastructure coordination

lampajr:
>Last but not least: how did you produces the chart? Which rate and what you have observed?

I created those charts with ~20 users per sec using a slightly complex scenario, I will try to create a very simple "hell-world"-like benchmark to highlight my findings. Worth to note is that it was mainly a personal impression and doubt.. so as you said the problem/lag was already there from the master code but I was thinking if with the previous change we were actually accentuate that lag or not.

>Hopefully, thanks to this change, what you have observed should be mitigated (fingers crossed), but I would like some test which we can put in for this, given that it can impact users and we need to keep it under control, for regression purposes.

Will take a look at this change asap :pray: 

>Regardless, I think we should improve the ElasticPool to:
> ...
>What I see is that we should have a jmh module to verify how the core classes behave and how much they cost, too.

I agree, both would be very useful but I would create separate issue for this as I would like to avoid mixing a lot of changes in this refactoring

>I've just noticed something odd: see https://github.com/Hyperfoil/Hyperfoil/blob/master/api/src/main/java/io/hyperfoil/api/connection/Request.java#L63 and https://github.com/Hyperfoil/Hyperfoil/blob/master/http/src/main/java/io/hyperfoil/http/steps/SendHttpRequestStep.java#L65: TLDR it seems to me that requests have no memory of the so called "intended" start time, inherited by the sessions run on the phase instance, which makes us prone to coordinated omission (!!) on master too.

Ouch, that's surprising...

>but the current solution, which can spin while computing the next fire time, just further increase the skew between the load generator fire time

yeah exactly, because we start all sessions/users only once we have computed all nextSession fire time (i.e., once we have computed how many users we should start). TBH I think that the most visible part is actually due to the fact that the very first user fire time is randomized as well.. thus the lag at phase startup.

core/src/main/java/io/hyperfoil/core/impl/user/PointPoissonConstantUserGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    protected double nextFireTimeMs(final double elapsedTimeMs) {
+        return fireTimeMs(random, usersPerSec, elapsedTimeMs);
+    }
+
+    static double fireTimeMs(final Random random, final double usersPerSec, final double elapsedTimeMs) {
+        // the Math.max(1e-20, random.nextDouble()) is to prevent the logarithm to be negative infinity, because
+        // the random number can approach to 0 and the logarithm of 0 is negative infinity
+        return Math.ceil(elapsedTimeMs + 1000 * -Math.log(Math.max(1e-20, random.nextDouble())) / usersPerSec);
+    }
```
lampajr, line 53: Given that the static method `fireTimeMs` is not called anywhere else, unless you are planning to do so I think we can remove it and move the logic directly inside `nextFireTimeMs`

```suggestion
    PointPoissonConstantUserGenerator(final double usersPerSec) {
        this.random = new Random();
        this.usersPerSec = usersPerSec;
        this.scheduledUser.fireTime = fireTimeMs(0);
    }

    @Override
    protected double nextFireTimeMs(final double elapsedTimeMs) {
        // the Math.max(1e-20, random.nextDouble()) is to prevent the logarithm to be negative infinity, because
        // the random number can approach to 0 and the logarithm of 0 is negative infinity
        return Math.ceil(elapsedTimeMs + 1000 * -Math.log(Math.max(1e-20, random.nextDouble())) / usersPerSec);
    }
```

core/src/main/java/io/hyperfoil/core/impl/OpenModelPhase.java
```diff
@@ -0,0 +1,101 @@
...
+        long elapsedTimeMs = fireTimeMs - absoluteStartTime;
+        var nextUser = userGenerator.scheduleNextUser(elapsedTimeMs);
+        startNewUsers(nextUser.users());
+        // we need to make sure that the scheduling decisions are made based on the current time
+        long schedulingMs = System.currentTimeMillis() - fireTimeMs;
+        // TODO: epoch from ms is still subject to NTP drifts: maybe we should do something here to account for it!
+        assert schedulingMs >= 0;
+        long scheduledFireTimeMs = nextUser.fireTime();
+        long scheduledFireDelayMs = (scheduledFireTimeMs - elapsedTimeMs) - schedulingMs;
+        // given that both scheduleNextUser and startNewUsers can take some time, we need to adjust the fire time
```
lampajr, line 60:
```suggestion
        // given that both scheduleNextUser and startNewUsers can take some time, we need to adjust the fire 
        long scheduledFireDelayMs = (nextUser.fireTime() - elapsedTimeMs) - schedulingMs;time
```

core/src/main/java/io/hyperfoil/core/impl/OpenModelPhase.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Override
+    public void proceed(final EventExecutorGroup executorGroup) {
+        if (status.isFinished()) {
+            return;
+        }
+        // TODO: epoch from ms is still subject to NTP drifts: maybe we should do something here to account for it!
+        long fireTimeMs = System.currentTimeMillis();
+        long elapsedTimeMs = fireTimeMs - absoluteStartTime;
+        var nextUser = userGenerator.scheduleNextUser(elapsedTimeMs);
```
lampajr, line 52: I am not sure I like the name `nextUser` as it might be confusing given that there are possibly more than one user scheduled. Maybe something like `nextScheduledSlot`.. I am not that good at giving names TBH :smile:

lampajr (commented):
Added a couple of other minor comments

core/src/main/java/io/hyperfoil/core/impl/OpenModelPhase.java
franz1981: I've added a new commit which rename the most of the rate generator classes but not the gigantic javadoc I made (I'll wait till the end for such).
Now the behaviour should be 1:1 with the previous one (or better)

franz1981:
I've happily found a weird bug on master thanks to this, so...keep on working on it

core/src/main/java/io/hyperfoil/core/impl/rate/FireTimeListener.java
```diff
@@ -0,0 +1,14 @@
...
+
+    void onMissedFireTime();
+
+    default void onMissedFireTimes(long count) {
+        for (long i = 0; i < count; i++) {
+            onMissedFireTime();
+        }
+    }
+
+}
```
lampajr, line 14: I agree that `onMissedFireTimes` is actually called as soon as we past the `elapsedTime` (i.e., as if we are always late) but this is actually how the logic was thought so calling `onMissedFireTimes` is the expected behavior, thus I would use a different name, e.g., `onFireTimes` and `onFireTime` and perhaps put the `missed` reasoning as comment just to highlight that this is expected. wdyt?

lampajr (commented):
I really like the new hierarchy and naming you used, I left just one comment on the `FireTimeListener`.

Just for the records, this is graphical view of the hierarchy:
```mermaid
graph TD
    RateGenerator --> BaseRateGenerator
    BaseRateGenerator --> FunctionalRateGenerator
    BaseRateGenerator --> SequentialRateGenerator
    FunctionalRateGenerator --> ConstantRateGenerator
    FunctionalRateGenerator --> RampRateGenerator
    SequentialRateGenerator --> PoissonConstantRateGenerator
    SequentialRateGenerator --> PoissonRampRateGenerator
```

franz1981:
I've added 1a480512a99804227c303701865086893e1a7657 and found something strange (again) i.e. the `Math::ceil` original behaviour was different from what I did, before 1a480512a99804227c303701865086893e1a7657 so with this last commit it should behave (apart from other adjustments), the same as the original code, but...
- if we compute the next fire time using Math::ceil, the next fire time will appear as happening later, although it is not eg double fire time = 0.7 long fire time (with Math::ceil) = 1
- given that it appear as been later and the original code loop by using a precise notion of time, it can take few iterations to keep up (and not just one, as expected) based on the distance applied by Math::ceil
- this can result in more injected fire times and Sessions

IDK if we should fix it somehow or is another non-problem, but I would like to make sure we're not regressing in some behaviour, hence the commit pasted above.

franz1981:
Tomorrow hopefully I'll add some statistical tests for the poisson part and I can undraft this.

franz1981:
I've been forced to add a `KolmogorovSmirnovTest` for the poisson distributions (which seems to work fine, yeeee), but I had to introduce nano seconds granularity on the generators for that: the skew introduced before was causing intervals to NOT follow the exponential distribution, making statistical tests, useless.

I have to clean this up and we should be ready to go (fingers crossed, is a huge change just to...fix few bugs and introduce proper testing + docs).

franz1981:
I've decided to NOT fix all possible problems I've found on the current rate generation implementation (one among many is #342), but this PR is delivering:

- :+1:  isolation among the different rate generation algorithms
- :+1: documented behaviours and the math/statistics behind each of them (especially the randomized ones)
- :+1: :-1:  same double/long approximations as before - to save introducing bugs not yet covered by tests
- :+1: :-1:  a whole new set of statistical tests - ignored due to the previous point 
- :-1: a fix for randomized loads which was causing to "spinning" (keep on submitting to different event loops) till the first fire time was exceeded
- :+1: a fix for late/skewed fire times for randomized loads which can further delays the next scheduled fire time
- :+1: minor optimizations precomputing the constants required to generated the load

The downside of this refactoring is the introduction of 4 different concrete classes which implements their own methods, while before was just 2 **but** the practical numbers of concrete classes required for a user test is still 2, as before, given that users usually doesn't mix randomized and non randomized rate generations in the same test; or better, this shouldn't be the common use case.

franz1981:
PTAL again @lampajr

lampajr (commented):
There is still something weird, I tried to run a very simple benchmark [1] using the constantRate and without variance

With a benchmark having 10 userPerSec that lasts 10secs, I would have expected 100 requests.. but I got 101. I checked the "as-is" behavior and the number of requests is as expected.

Before
image

After
image

I tried with 3 `userPerSec` and lower values and the issue does not apply.. with more than 3, it does instead:
image
image

I will deep dive into this in a while.

[1] https://hyperfoil.io/2021/01/25/hf-beginner-guide-1.html#first-benchmark

lampajr:
> There is still something weird, I tried to run a very simple benchmark [1] using the constantRate and without variance
> 
> With a benchmark having 10 userPerSec that lasts 10secs, I would have expected 100 requests.. but I got 101. I checked the "as-is" behavior and the number of requests is as expected.
> I will deep dive into this in a while.

Ok I checked a bit more the behavior and:
- It looks like it depends when the last fire time is scheduled, if it matches the exact phase termination it might fire an additional user:
```
11:34:52,900 WARN  (epollEventLoopGroup-3-1) [i.h.c.i.PhaseInstanceImpl] fetchIndex: 10000 after start, 41 started (0 throttled), next user in 250 ms, scheduling decisions tooks 0 ms
```
- This is not deterministic
- Most important: most likely this has not been introduced by this refactoring as I was able to reproduce the same behavior in the actual `master` code.

franz1981:
PTAL @lampajr

IDK if i should add an end 2 end test which measure the amount of started sessions, because will likely be broken till #368 will be fixed

lampajr (approved):
>IDK if i should add an end 2 end test which measure the amount of started sessions, because will likely be broken till https://github.com/Hyperfoil/Hyperfoil/issues/368 will be fixed

I think this is really enough at the moment, I did not spot anything else that could be suspicious.

That was a huge work @franz1981 :clap:

franz1981:
@lampajr if we're fine I can merge it? or any else you need to verify?

lampajr:
> @lampajr if we're fine I can merge it? or any else you need to verify?

Already merged! :rocket:

PR #367, opened 2024-06-20
Uses JCTools primary lock-free queue on Session pooling

### LINKED ISSUE #359: Improve Session pool
The comments at https://github.com/Hyperfoil/Hyperfoil/blob/hyperfoil-all-0.26/core/src/main/java/io/hyperfoil/core/impl/ElasticPoolImpl.java#L27-L28 and https://github.com/Hyperfoil/Hyperfoil/pull/357#issuecomment-2156186667 shows that we have different ways to improve the `ElasticPool` powering the Session's pool.

The easiest solution has been attempted at https://github.com/franz1981/Hyperfoil/commit/25948a82e02ec30e09cd8f5e3b96f14ee4279c50 which should already improve the current situation, but we need a jmh module to prove it.

franz1981:
Fixes #359 

## Changes proposed
- [x] Use a lock-free mpmc queue for the primary queue's elastic pool used for Session pooling
- [ ] JCTools mpmc unbounded xadd queue comes with a footprint cost: evaluate if is better to just use CLQ   
- [ ]  Introduce a shiny new JMH module where we can put new benchmarks, including something to validate the previous point 
- [x] Improve `reserve` to recycle the already pooled sessions while expanding (it affect `wrk2`) 
- [ ]  write a test which reproduce the above the problem
- [ ] evaluate if the pool should be elastic or not, given that AFAIK we cannot create sessions on demand and we use it just for sessions

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -66,6 +66,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.jctools</groupId>
+            <artifactId>jctools-core</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>io.netty</groupId>
             <artifactId>netty-tcnative-boringssl-static</artifactId>
         </dependency>
--- a/core/src/main/java/io/hyperfoil/core/impl/ElasticPoolImpl.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/ElasticPoolImpl.java
@@ -1,12 +1,15 @@
 package io.hyperfoil.core.impl;
 
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.LinkedBlockingQueue;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.LongAdder;
 import java.util.function.Supplier;
 
 import io.hyperfoil.api.collection.ElasticPool;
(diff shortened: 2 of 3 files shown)


franz1981:
After some time benchmarking it I admit I am not impressed by the performance of the JCTools queues, but I shouldn't be surprised because https://github.com/JCTools/JCTools/pull/374 i have already performed this same analysis :"(

I have an idea of using striped fixed number of mpmc queues (matching the number of event loops) which are assigned in round robin via fast thread locals to the acquiring threads: each stripe can record the amount of items drained by the non assigned queues in order to make the paired release (from the same acquiring Thread) able to return back the items to the right queue.
These additional counters can be just plain integers, because are always accessed by the same thread.
This would guarantee in the optimal case to have an exclusive access to the dedicated queue while spreading the contention among the others in the slow path case.
We still uses queues here despite not needed, and mostly because are GC free.

franz1981:
After an experiment on https://github.com/franz1981/Hyperfoil/tree/rate_generators_session_pool_workstealing I've found an issue with the provided solutions: we need to track the origin of a session slot or we won' be able to cover for the "typical" use case for hyperfoil:

1. the rate generator acquire a session from event loop X
2. the same rate generator release the same session from the Session's executor

In short: the session (or the pool) should remember where the pooled items belong - which is way easier if we leave the `SessionImpl` able to do it.

franz1981:
This is closed in favour of https://github.com/Hyperfoil/Hyperfoil/pull/374

PR #369, opened 2024-06-21
[cli/stats] Add max response time

wfouche:
Max response time is one of the most important response time values to keep an eye on, and the stats display already had enough percentile response time values.

````
[hyperfoil@in-vm]$ stats
Total stats from run 0004
PHASE    METRIC  THROUGHPUT   REQUESTS  MEAN      STD_DEV  p50       p90       p99       p99.9     MAX       TIMEOUTS  ERRORS  BLOCKED  2xx  3xx  4xx  5xx  CACHE
example  test    29,41 req/s         1  17,37 ms     0 ns  17,43 ms  17,43 ms  17,43 ms  17,43 ms  17,43 ms         0       0     0 ns    0    1    0    0      0
[hyperfoil@in-vm]$ 
````

### DIFF
--- a/cli/src/main/java/io/hyperfoil/cli/commands/Stats.java
+++ b/cli/src/main/java/io/hyperfoil/cli/commands/Stats.java
@@ -34,7 +34,7 @@ public class Stats extends BaseRunIdCommand {
          .columnNanos("p90", r -> r.summary.percentileResponseTime.get(90d))
          .columnNanos("p99", r -> r.summary.percentileResponseTime.get(99d))
          .columnNanos("p99.9", r -> r.summary.percentileResponseTime.get(99.9))
-         .columnNanos("p99.99", r -> r.summary.percentileResponseTime.get(99.99))
+         .columnNanos("MAX", r -> r.summary.maxResponseTime)
          .columnInt("TIMEOUTS", r -> r.summary.requestTimeouts)
          .columnInt("ERRORS", r -> r.summary.connectionErrors + r.summary.internalErrors)
          .columnNanos("BLOCKED", r -> r.summary.blockedTime);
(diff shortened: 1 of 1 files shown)


franz1981:
I agree @wfouche , but max risk to capture systematic latencies/hiccups too, which, despite being important, are not always related what Hyperfoil/nor users can do about it, while holding on a bit (i.e. the so called p99.99 one)  is still informative without dealing with system noise/hiccups.

Said that, the cli is not meant to replace the full response time histograms, which contains a more detailed view, but I see no harm into adding max there, as long as it turn to be valuable: and having p99.99 as well in, help to distinguish between systematic hiccups vs actual useful data.

wfouche:
Thanks for the feedback @franz1981

I see the System Under Test as a musical instrument that is being tuned and Hyperfoil as a tuning fork.  :-)

Worst than expected max response times are "dissonant notes" that one should be aware of.  So I think having MAX response time displayed as well is extremely useful. 

I'll update the pull request and add p99.99 back, and will also move MAX response time to be next to and to the right of STD_DEV.

````
PHASE    METRIC  THROUGHPUT   REQUESTS  MEAN      STD_DEV  MAX       p50       p90       p99       p99.9     p99.99    TIMEOUTS  ERRORS  BLOCKED  2xx  3xx  4xx  5xx  CACHE
example  test    30,30 req/s         1  11,17 ms     0 ns  11,21 ms  11,21 ms  11,21 ms  11,21 ms  11,21 ms  11,21 ms         0       0     0 ns    0    1    0    0      0
````

lampajr (approved):
LGTM!

I am just thinking if we would like to expose the same metric in the report or not :thinking:  but I think we can go with this anyway

Issue #371, opened 2024-06-25
Add some additional Hyperfoil internal time metrics


lampajr:
Add some time metrics:
- [ ] Parsing time: time spent parsing/creating the benchmark in memory
- [ ] Startup time: time taken by the application before actually starting the phase (this will include the overall java allocation)

lampajr:
After taking a look a bit more, I don't see a real value in adding the parsing time at the moment as it is actually out of the hot path.

When you run a benchmark, the corresponding object is already in memory (properly parsed) as the parsing has been already performed as part of the benchmark `upload`.

PR #372, opened 2024-06-25
Add simulation initialization time log

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Hyperfoil/issues/371

## Changes proposed

Add simple time log identifying the time spent by the simulation initialization, which includes the sessions allocation and `GC` manual invocation.

As part of this monitoring enhancement, the following changes
have been added:
- Overall simulation init execution time
- GC execution time
- Run GC twice with forced finalization
- Additional check to ensure GC properly worked (configurable)

The additional check is disabled by default but it can be
switched on by providing -Dio.hyperfoil.gc.check.enabled=true

<details>
<summary>
</summary>

</details>

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/impl/SimulationRunner.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/SimulationRunner.java
@@ -1,5 +1,7 @@
 package io.hyperfoil.core.impl;
 
+import java.lang.management.GarbageCollectorMXBean;
+import java.lang.management.ManagementFactory;
 import java.text.SimpleDateFormat;
 import java.time.Clock;
 import java.util.ArrayList;
@@ -99,6 +101,7 @@ public class SimulationRunner {
    }
 
    public void init() {
+      long initSimulationStartTime = System.currentTimeMillis();
       AgentData agentData = new AgentDataImpl();
       ThreadData[] threadData = new ThreadData[executors.length];
       Arrays.setAll(threadData, executorId -> new ThreadDataImpl());
@@ -150,13 +153,15 @@ public class SimulationRunner {
          // at this point all session resources should be reserved
       }
       // hint the GC to tenure sessions
-      System.gc();
+      this.runGC();
 
       jitterWatchdog = new Thread(this::observeJitter, "jitter-watchdog");
       jitterWatchdog.setDaemon(true);
 
       cpuWatchdog = new CpuWatchdog(errorHandler, () -> instances.values().stream().anyMatch(p -> !p.definition().isWarmup));
       cpuWatchdog.start();
+
+      log.info("Simulation initialization took {} ms", System.currentTimeMillis() - initSimulationStartTime);
    }
 
    public void openConnections(Function<Callable<Void>, Future<Void>> blockingHandler, Handler<AsyncResult<Void>> handler) {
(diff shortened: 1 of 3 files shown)


franz1981:
Despite the GC being an hint I have a proposal for you to have a separate and let's say "reliable" measure of the portion of time spent exclusively on GC collection...wait a sec that I found the relevant code to link

franz1981:
That's it: https://github.com/openjdk/jmh/blob/6d6ce6315dc39d1d3abd0e3ac9eca9c38f767112/jmh-core/src/main/java/org/openjdk/jmh/runner/BaseRunner.java#L309

We should think to have something like this which takes care of GC to happen "for real", but we meed to prove this to not harm what we got

lampajr:
> That's it: https://github.com/openjdk/jmh/blob/6d6ce6315dc39d1d3abd0e3ac9eca9c38f767112/jmh-core/src/main/java/org/openjdk/jmh/runner/BaseRunner.java#L309
> 
> We should think to have something like this which takes care of GC to happen "for real", but we meed to prove this to not harm what we got

Interesting, so instead of simply calling `System.gc` we will wrap it as ~done in the link you provided. I will work on it :muscle:

franz1981:
yep - it should have a limit of 20 ms of whole duration - and allocate something, but it has the beauty of make sure that GC is being performed for real - and actual young gen object will likely goes into old gen.
Additionally it warn if something is not working as expected.
Having this duration enable us to understand the actual time to perform this operation and will help us to understand, without profiling, if is a too costy operation or if we have too much allocations or live set in the test.
Additionally you can make use of this to verify that your changes to remove the HTTP cache are effective in the 2 phases:
- before GC start
- GC duration

which are both dependent by how much live set we got.

lampajr:
I've been able to implement a configurale GC monitoring, by default it is disabled but it can be turned on providing `-D-Dio.hyperfoil.gc.check.enabled=true`.

@franz1981 I kept running the GC cycle twice internally as IIUC this seems widely used to clear unreachable objects that were not identified during the first GC cycle. This call also triggers finalization to release resources that were used by the unreachable objects so that the resources can be reused. Correct me if I am wrong.

core/src/main/java/io/hyperfoil/core/impl/SimulationRunner.java
```diff
@@ -411,6 +416,119 @@ public void addGlobalData(Map<String, GlobalData.Element> globalData) {
...
+         // Make sure the GC actually happened
+         //   a) That at least two collections happened, indicating GC work.
+         //   b) That counter updates have not happened for a while, indicating GC work had ceased.
+
+         final int MAX_WAIT_MS = 10 * 1000;
+
+         if (enabledBeans.isEmpty()) {
+            log.warn("MXBeans can not report GC info. System.gc() invoked, pessimistically waiting {} ms", MAX_WAIT_MS);
+            try {
+               TimeUnit.MILLISECONDS.sleep(MAX_WAIT_MS);
```
franz1981, line 460: We could failback to the original behaviour here, instead of letting the user to wait for so long

franz1981 (changes requested):
LGTM just minor thing commented

core/src/main/java/io/hyperfoil/core/impl/SimulationRunner.java
lampajr: I agree, fixed :heavy_check_mark: 

I also switched the if-else as you were suggesting offline such that the default is the first one

core/src/main/java/io/hyperfoil/core/impl/SimulationRunner.java
```diff
@@ -411,6 +416,117 @@ public void addGlobalData(Map<String, GlobalData.Element> globalData) {
...
+   private void runSystemGC() {
+      long actualGCRun = System.currentTimeMillis();
+
+      // Run the GC twice and force finalization before every GC run
+      System.runFinalization();
+      System.gc();
+      System.runFinalization();
+      System.gc();
+
+      log.debug("GC execution took {} ms", (System.currentTimeMillis() - actualGCRun));
```
franz1981, line 527: Why debug?

lampajr: Good catch, it was a leftover.. fixed now!

franz1981 (approved):
LGTM, well done @lampajr !

PR #374, opened 2024-06-26
Rate generators session pool workstealing

### LINKED ISSUE #359: Improve Session pool

franz1981:
Fixes #359 

## Changes proposed

- [x] Implemented basic striped pool implementation
- [X] Adding unit tests
- [X] Adding JMH benchmark module
- [x] Leverage Thread-affinity of `Session`s to populate and assign it to dedicated queues

The last point is crucial to better work with the phase instance which acquire `Session`s: instead of relying on a random thread affinity (assigned in round robin) - we should just use the initial provided one (thinking about the `Session`s specific use case here) in order to both reduce contention and guarantee the thread-local acquired `Session` is executed in the same `Thread` who requested it.
The implication of this choice are not obvious because it means that in a multithreading scenario where the phase instance acquire the Session, it will start it delayed after it complete the `proceed` call, which is something to be evaluated if is the right choice.

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/impl/AffinityAwareSessionPool.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/AffinityAwareSessionPool.java
@@ -1,0 +1,258 @@
+package io.hyperfoil.core.impl;
+
+import io.hyperfoil.api.collection.ElasticPool;
+import io.hyperfoil.api.session.Session;
+import io.netty.util.concurrent.EventExecutor;
+import io.netty.util.concurrent.FastThreadLocal;
+import io.netty.util.internal.PlatformDependent;
+import io.netty.util.internal.ThreadExecutorMap;
+import org.jctools.queues.MpmcArrayQueue;
+import org.jctools.queues.atomic.MpmcAtomicArrayQueue;
+
+import java.util.IdentityHashMap;
+import java.util.Objects;
+import java.util.Queue;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicIntegerArray;
+import java.util.function.Supplier;
+
+/**
+ * This class represents a pool of sessions that are aware of thread affinity.
+ * It implements the {@link ElasticPool} interface and provides methods to acquire and release sessions.<br>
+ * The pool maintains a separate queue of sessions for each event executor (thread),
+ * and tries to acquire and release sessions from the queue of the current executor.
+ * If the local queue is empty, it tries to steal sessions from other queues.
+ */
+public class AffinityAwareSessionPool implements ElasticPool<Session> {
+
+    private final FastThreadLocal<Integer> localAgentThreadId;
+    private final IdentityHashMap<EventExecutor, Integer> agentThreadIdPerExecutor;
+    private final Queue<Session>[] localQueues;
+    private final Supplier<Session> sessionSupplier;
+
+    /**
+     * We're using a single array to store multiple counters, mostly to avoid false sharing with other fields
+     * in this same class (which are mostly read-only).<br>
+     * We still pad used/min/max from each others to avoid max and min to interfere with each other.
+     * There's no point to use a LongAdder here, because reading the counters would hit worse than sharing a single one
+     * which can leverage fast x86 fetch-and-add instructions.
+     */
+    private static final int PADDING_INTS = 128 / Integer.BYTES; // 32
+    private static final int USED_OFFSET = PADDING_INTS; // 32
+    private static final int MIN_USED_OFFSET = USED_OFFSET + PADDING_INTS; // 32 + 32 = 64
+    private static final int MAX_USED_OFFSET = MIN_USED_OFFSET + PADDING_INTS; // 64 + 32 = 96
+    private static final int COUNTERS_INTS = MAX_USED_OFFSET + PADDING_INTS; // 96 + 32 = 128
+    private final AtomicIntegerArray counters;
+
+    public AffinityAwareSessionPool(EventExecutor[] eventExecutors, Supplier<Session> sessionSupplier) {
+        this.sessionSupplier = sessionSupplier;
+        this.agentThreadIdPerExecutor = new IdentityHashMap<>(eventExecutors.length);
+        this.localQueues = new Queue[eventExecutors.length];
+        for (int agentThreadId = 0; agentThreadId < eventExecutors.length; agentThreadId++) {
+            agentThreadIdPerExecutor.put(eventExecutors[agentThreadId], agentThreadId);
+        }
+        this.localAgentThreadId = new FastThreadLocal<>() {
+            @Override
+            protected Integer initialValue() {
+                var eventExecutor = ThreadExecutorMap.currentExecutor();
+                return eventExecutor == null ? null : agentThreadIdPerExecutor.get(eventExecutor);
+            }
+        };
+        this.counters = new AtomicIntegerArray(COUNTERS_INTS);
+    }
+
+    private int nextWorkStealIndex() {
+        return ThreadLocalRandom.current().nextInt(0, localQueues.length);
+    }
+
+    private int nextWorkStealIndex(int excludeIndex) {
+        var localQueuesCount = localQueues.length;
+        var workStealIndex = localQueuesCount == 2 ? excludeIndex : ThreadLocalRandom.current().nextInt(0, localQueuesCount);
+        if (workStealIndex == excludeIndex) {
+            workStealIndex++;
+            if (workStealIndex == localQueuesCount) {
+                workStealIndex = 0;
+            }
+        }
+        return workStealIndex;
+    }
+
+
+    @Override
+    public Session acquire() {
+        var boxedAgentThreadId = localAgentThreadId.get();
+        if (boxedAgentThreadId == null) {
+            return acquireFromLocalQueues();
+        }
+        // explicitly unbox to save unboxing it again, in case work-stealing happens
+        var agentThreadId = boxedAgentThreadId.intValue();
+        var localQueues = this.localQueues;
+        var session = localQueues[agentThreadId].poll();
+        if (session != null) {
+            incrementUsed();
+            return session;
+        }
+        if (localQueues.length == 1) {
+            return null;
+        }
+        return acquireFromOtherLocalQueues(localQueues, agentThreadId);
+    }
+
+    private Session acquireFromLocalQueues() {
+        int currentIndex = nextWorkStealIndex();
+        var localQueues = this.localQueues;
+        int localQueuesCount = localQueues.length;
+        for (Queue<Session> localQueue : localQueues) {
+            Session session = localQueue.poll();
+            if (session != null) {
+                incrementUsed();
+                return session;
+            }
+            currentIndex++;
+            if (currentIndex == localQueuesCount) {
+                currentIndex = 0;
+            }
+        }
+        return null;
+    }
+
+    private Session acquireFromOtherLocalQueues(Queue<Session>[] localQueues, int agentThreadIdToSkip) {
+        int currentIndex = nextWorkStealIndex(agentThreadIdToSkip);
+        var localQueuesCount = localQueues.length;
+        // this loop start from the workStealIndex, but while iterating can still end up
+        // at the index we want to ignore, so we need to skip it
+        for (int i = 0; i < localQueuesCount; i++) {
+            if (currentIndex != agentThreadIdToSkip) {
+                var localQ = localQueues[currentIndex];
+                if (localQ != null) {
+                    var session = localQ.poll();
+                    if (session != null) {
+                        incrementUsed();
+                        return session;
+                    }
+                }
+            }
+            currentIndex++;
+            if (currentIndex == localQueuesCount) {
+                currentIndex = 0;
+            }
+        }
+        return null;
+    }
+
+    private void incrementUsed() {
+        var counters = this.counters;
+        int used = counters.incrementAndGet(USED_OFFSET);
+        int maxUsed = counters.get(MAX_USED_OFFSET);
+        if (used > maxUsed) {
+            counters.lazySet(MAX_USED_OFFSET, used);
+        }
+    }
+
+    private void decrementUsed() {
+        var counters = this.counters;
+        int used = counters.decrementAndGet(USED_OFFSET);
+        int minUsed = counters.get(MIN_USED_OFFSET);
+        if (minUsed > 0 && used < minUsed) {
+            counters.lazySet(MIN_USED_OFFSET, used);
+        }
+    }
+
+    @Override
+    public void release(Session session) {
+        Objects.requireNonNull(session);
+        var localQueue = localQueues[session.agentThreadId()];
+        decrementUsed();
+        localQueue.add(session);
+    }
+
+    @Override
+    public void reserve(int capacity) {
+        int totalCapacity = getLocalQueuesCapacity();
+        if (totalCapacity >= capacity) {
+            return;
+        }
+        moveNewSessionsToLocalQueues(capacity, totalCapacity);
+    }
+
+    private int getLocalQueuesCapacity() {
+        int totalCapacity = 0;
+        for (Queue<Session> localQueue : localQueues) {
+            if (localQueue != null) {
+                totalCapacity += localQueue.size();
+            }
+        }
+        return totalCapacity;
+    }
+
+    private void moveNewSessionsToLocalQueues(int requiredCapacity, int currentCapacity) {
+        final int newCapacity = requiredCapacity - currentCapacity;
+        // assume fair distribution of new sessions
+        var perEventExecutorNewCapacity = (int) Math.ceil((double) newCapacity / localQueues.length);
+        var sessionSupplier = this.sessionSupplier;
+        var localQueues = this.localQueues;
+        var agentThreadIdPerExecutor = this.agentThreadIdPerExecutor;
+        // It keeps track of the local queues capacity not yet reserved
+        boolean[] localQueueReservedCapacity = new boolean[localQueues.length];
+        for (int i = 0; i < newCapacity; i++) {
+            var newSession = sessionSupplier.get();
+            var eventExecutor = newSession.executor();
+            var boxedAgentThreadId = agentThreadIdPerExecutor.get(eventExecutor);
+            if (boxedAgentThreadId == null) {
+                throw new IllegalStateException("No agentThreadId for executor " + eventExecutor);
+            }
+            var agentThreadId = boxedAgentThreadId.intValue();
+            var localQueue = localQueues[agentThreadId];
+            if (!localQueueReservedCapacity[agentThreadId]) {
+                localQueueReservedCapacity[agentThreadId] = true;
+                if (localQueue == null) {
+                    localQueue = createLocalQueue(perEventExecutorNewCapacity);
+                    localQueues[agentThreadId] = localQueue;
+                } else {
+                    var newLocalQueue = createLocalQueue(localQueue.size() + perEventExecutorNewCapacity);
+                    newLocalQueue.addAll(localQueue);
+                    localQueue.clear();
+                    localQueue = newLocalQueue;
+                    localQueues[agentThreadId] = newLocalQueue;
+                }
+            }
+            if (!localQueue.offer(newSession)) {
+                throw new IllegalStateException("Failed to add new session to local queue: sessions are not fairly distributed");
+            }
+        }
+        for (int i = 0; i < localQueueReservedCapacity.length; i++) {
+            if (!localQueueReservedCapacity[i]) {
+                var localQ = localQueues[i];
+                if (localQ == null) {
+                    localQ = createLocalQueue(0);
+                    this.localQueues[i] = localQ;
+                }
+            }
+        }
+    }
+
+    private Queue<Session> createLocalQueue(int capacity) {
+        if (PlatformDependent.hasUnsafe()) {
+            return new MpmcArrayQueue<>(Math.max(2, capacity));
+        }
+        return new MpmcAtomicArrayQueue<>(Math.max(2, capacity));
+    }
+
+    @Override
+    public int minUsed() {
+        return counters.get(MIN_USED_OFFSET);
+    }
+
+    @Override
+    public int maxUsed() {
+        return counters.get(MAX_USED_OFFSET);
+    }
+
+    @Override
+    public void resetStats() {
+        var counters = this.counters;
+        int used = counters.get(USED_OFFSET);
+        counters.lazySet(MAX_USED_OFFSET, used);
+        counters.lazySet(MIN_USED_OFFSET, used);
+    }
+}
--- a/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
@@ -40,6 +40,7 @@ public abstract class PhaseInstanceImpl implements PhaseInstance {
    protected AtomicInteger activeSessions = new AtomicInteger(0);
    private volatile Throwable error;
    private volatile boolean sessionLimitExceeded;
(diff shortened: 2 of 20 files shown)


franz1981:
I am not really happy about 4d740c3efe2e7683335078c7997b5ce350a6243c (Which is hanging the CI) mostly because of a quick analysis:
1. phase instances start seems to assume it runs immediately 
2. Such method is not always called from event loops (on simulation start)
3. It breaks an event loop confined pool like the one in this pr 
4. Forcing it to start on the event loop just breaks the CI with test failures - which get several NPE here and there

I have to dig into the root cause and if the assumption on start is there for some reason, or just relax the new pool requirements - but is a shame because it is for a single acquire request.

franz1981:
I've still to improve the `reserve` performance and allocation rate (by far, actually!)
I've added a microbenchmark just for it, which I still have to analyze and improve, but code-wise this is finally ready for review

core/src/main/java/io/hyperfoil/core/impl/AffinityAwareSessionPool.java
```diff
@@ -0,0 +1,271 @@
...
+        var localQueues = this.localQueues;
+        var session = localQueues[agentThreadId].poll();
+        if (session != null) {
+            incrementUsed();
+            return session;
+        }
+        if (localQueues.length == 1) {
+            return null;
+        }
+        return acquireFromOtherLocalQueues(localQueues, agentThreadId);
```
lampajr, line 107: Couple of questions here:
- Why do we need explicit unboxing for `boxedAgentThreadId`? isn't it already automatically managed?
- Why do we need to save `this.localQueues` to a local variable? is there any particular reason or benefit?

If there is no particular reason fo rboth doubts I think we can simplify a little bit:
```suggestion
        var boxedAgentThreadId = localAgentThreadId.get();
        if (boxedAgentThreadId == null) {
            return acquireFromLocalQueues();
        }
        var session = localQueues[boxedAgentThreadId].poll();
        if (session != null) {
            incrementUsed();
            return session;
        }
        if (this.localQueues.length == 1) {
            return null;
        }
        return acquireFromOtherLocalQueues(this.localQueues, boxedAgentThreadId);
```

core/src/main/java/io/hyperfoil/core/impl/AffinityAwareSessionPool.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    private Session acquireFromLocalQueues() {
+        int workStealIndex = nextWorkStealIndex();
+        int currentIndex = workStealIndex;
```
lampajr, line 112:
```suggestion
        int currentIndex = nextWorkStealIndex();
```

core/src/main/java/io/hyperfoil/core/impl/AffinityAwareSessionPool.java
```diff
@@ -0,0 +1,271 @@
...
+            if (newSession == null) {
+                throw new IllegalStateException("Session supplier returned null");
+            }
+            var executor = newSession.executor();
+            var boxedAgentThreadId = agentThreadIdPerExecutor.get(executor);
+            if (boxedAgentThreadId == null) {
+                throw new IllegalStateException("This session is associated with an unknown EventExecutor");
+            }
+            // sanity check against the sessions agentThreadId
+            if (boxedAgentThreadId.intValue() != newSession.agentThreadId()) {
```
lampajr, line 229:
```suggestion
            if (boxedAgentThreadId != newSession.agentThreadId()) {
```

lampajr (commented):
Hey @franz1981 this looks really great!

I did a first round and I found just some minors/simplification. I'll let you decide whether to include them or not.

I see you explicitly box/unbox integers, is there any particular reason for doing that?

core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
```diff
@@ -264,13 +265,23 @@ protected boolean startNewSession() {
...
       return true;
    }
 
+   private void noSessionsAvailable() {
+      if (failedSessionAcquisitionAction != null) {
+          failedSessionAcquisitionAction.run();
+      }
+      if (!sessionLimitExceeded) {
+         sessionLimitExceeded = true;
+      }
```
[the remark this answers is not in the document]
franz1981: The problem is that it is a costy operation to volatile set (it waste like hundreds of cycles more than a "plain" non volatile set") - so I suppose that Radim decided to pay this just the first time (unless reset happen)

core/src/main/java/io/hyperfoil/core/impl/AffinityAwareSessionPool.java
franz1981: The main reason was to avoid twice `Integer::intValue`: given that there are atomic operations in-between the 2 unboxing operations, the JIT can decide to actual perform it twice, without trusting that Integer's value hasn't changed :/

core/src/main/java/io/hyperfoil/core/impl/AffinityAwareSessionPool.java
franz1981: I don't remember by heart if the left side in unboxed or the right side is boxed, so I preferred to be safe and avoid a potential allocation path.

core/src/main/java/io/hyperfoil/core/impl/AffinityAwareSessionPool.java
```diff
@@ -0,0 +1,271 @@
...
+                throw new IllegalStateException("Session::getAgentThreadId is not associated with the provided EventExecutor");
+            }
+            var agentThreadId = boxedAgentThreadId.intValue();
+            if (newSessionsPerEventExecutors[agentThreadId] == null) {
+                newSessionsPerEventExecutors[agentThreadId] = new ArrayDeque<>(perEventExecutorCapacity);
+            }
+            newSessionsPerEventExecutors[agentThreadId].add(newSession);
```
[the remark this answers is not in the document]
franz1981: reserving usually happen at the very begining of the application: i've saved unboxing to happen few times while the code is still cold, to save the potential `invokespecial` bytecode to unbox, to happen 3 more times, per session created

core/src/main/java/io/hyperfoil/core/impl/AffinityAwareSessionPool.java
franz1981: thanks, it was a leftover of the previous implamentation, good catch!

franz1981:
@lampajr I've yet to improve the reserve benchmark because I'm paying the class loading cost there :/ 
I have to try wrk2 (which uses a LOT of sessions, for high rates) and see if the diff in GC time is measurable at startup...
Right now given that we uses different queues (which are way bigger(s)) than `ArrayBlockingQueue` it is expected, but don't want to push this that far - just need to make sure that is not too much impacted.

franz1981:
```
$ java -jar target/benchmarks.jar PoolBenchmark -pworkStealingPool=false,true -pcapacity=1024 -t 1 -pfakeEventExecutors=0

Benchmark                                 (burst)  (capacity)  (fakeEventExecutors)  (work)  (workStealingPool)   Mode  Cnt   Score   Error   Units
PoolBenchmark.acquireAndRelease                 1        1024                     0       0               false  thrpt   10  11.673 ± 0.013  ops/us
PoolBenchmark.acquireAndRelease:acquired        1        1024                     0       0               false  thrpt   10  11.673 ± 0.013  ops/us
PoolBenchmark.acquireAndRelease:released        1        1024                     0       0               false  thrpt   10  11.673 ± 0.013  ops/us
PoolBenchmark.acquireAndRelease                 1        1024                     0       0                true  thrpt   10  17.461 ± 0.065  ops/us
PoolBenchmark.acquireAndRelease:acquired        1        1024                     0       0                true  thrpt   10  17.461 ± 0.065  ops/us
PoolBenchmark.acquireAndRelease:released        1        1024                     0       0                true  thrpt   10  17.461 ± 0.065  ops/us
```
```
$ java -jar target/benchmarks.jar PoolBenchmark -pworkStealingPool=false,true -pcapacity=1024 -t 2 -pfakeEventExecutors=0

Benchmark                                 (burst)  (capacity)  (fakeEventExecutors)  (work)  (workStealingPool)   Mode  Cnt  Score   Error   Units
PoolBenchmark.acquireAndRelease                 1        1024                     0       0               false  thrpt   10  2.628 ± 0.082  ops/us
PoolBenchmark.acquireAndRelease:acquired        1        1024                     0       0               false  thrpt   10  2.629 ± 0.082  ops/us
PoolBenchmark.acquireAndRelease:released        1        1024                     0       0               false  thrpt   10  2.629 ± 0.082  ops/us
PoolBenchmark.acquireAndRelease                 1        1024                     0       0                true  thrpt   10  9.518 ± 0.279  ops/us
PoolBenchmark.acquireAndRelease:acquired        1        1024                     0       0                true  thrpt   10  9.518 ± 0.279  ops/us
PoolBenchmark.acquireAndRelease:released        1        1024                     0       0                true  thrpt   10  9.518 ± 0.279  ops/us
```
```
$ java -jar target/benchmarks.jar PoolBenchmark -pworkStealingPool=false,true -pcapacity=1024 -t 4 -pfakeEventExecutors=0

Benchmark                                 (burst)  (capacity)  (fakeEventExecutors)  (work)  (workStealingPool)   Mode  Cnt  Score   Error   Units
PoolBenchmark.acquireAndRelease                 1        1024                     0       0               false  thrpt   10  6.461 ± 0.253  ops/us
PoolBenchmark.acquireAndRelease:acquired        1        1024                     0       0               false  thrpt   10  6.461 ± 0.253  ops/us
PoolBenchmark.acquireAndRelease:released        1        1024                     0       0               false  thrpt   10  6.461 ± 0.253  ops/us
PoolBenchmark.acquireAndRelease                 1        1024                     0       0                true  thrpt   10  8.513 ± 0.190  ops/us
PoolBenchmark.acquireAndRelease:acquired        1        1024                     0       0                true  thrpt   10  8.514 ± 0.190  ops/us
PoolBenchmark.acquireAndRelease:released        1        1024                     0       0                true  thrpt   10  8.514 ± 0.190  ops/us
```
The next benchmark run is obtained by configuring 8 event executors (4 real + 4 fake) which fairly distribute the configured capacity i.e. 128 / 8 = 16 sessions per event executor
The real threads which acquire/release are just 4, acquiring/releasing 32 sessions per benchmark invocation, which means that 16 of these sessions will belong to the local session while the remaining (16) are acquired from the other queues based on the workstealing policy:
```
$ java -jar target/benchmarks.jar PoolBenchmark -pworkStealingPool=false,true -pcapacity=128 -pburst=32 -t 4 -pfakeEventExecutors=4

Benchmark                                 (burst)  (capacity)  (fakeEventExecutors)  (work)  (workStealingPool)   Mode  Cnt  Score   Error   Units
PoolBenchmark.acquireAndRelease                32         128                     4       0               false  thrpt   10  0.205 ± 0.002  ops/us
PoolBenchmark.acquireAndRelease:acquired       32         128                     4       0               false  thrpt   10  6.575 ± 0.063  ops/us
PoolBenchmark.acquireAndRelease:released       32         128                     4       0               false  thrpt   10  6.575 ± 0.063  ops/us
PoolBenchmark.acquireAndRelease                32         128                     4       0                true  thrpt   10  0.205 ± 0.005  ops/us
PoolBenchmark.acquireAndRelease:acquired       32         128                     4       0                true  thrpt   10  6.577 ± 0.150  ops/us
PoolBenchmark.acquireAndRelease:released       32         128                     4       0                true  thrpt   10  6.577 ± 0.150  ops/us
```
Removing the fake event executors, instead (which means no workstealing is happening):
```
Benchmark                                 (burst)  (capacity)  (fakeEventExecutors)  (work)  (workStealingPool)   Mode  Cnt  Score   Error   Units
PoolBenchmark.acquireAndRelease                32         128                     0       0               false  thrpt   10  0.202 ± 0.004  ops/us
PoolBenchmark.acquireAndRelease:acquired       32         128                     0       0               false  thrpt   10  6.449 ± 0.135  ops/us
PoolBenchmark.acquireAndRelease:released       32         128                     0       0               false  thrpt   10  6.449 ± 0.135  ops/us
PoolBenchmark.acquireAndRelease                32         128                     0       0                true  thrpt   10  0.280 ± 0.059  ops/us
PoolBenchmark.acquireAndRelease:acquired       32         128                     0       0                true  thrpt   10  8.969 ± 1.900  ops/us
PoolBenchmark.acquireAndRelease:released       32         128                     0       0                true  thrpt   10  8.969 ± 1.900  ops/us
PoolBenchmark.acquireAndRelease                32         128                    16       0               false  thrpt   10  0.210 ± 0.002  ops/us
PoolBenchmark.acquireAndRelease:acquired       32         128                    16       0               false  thrpt   10  6.713 ± 0.057  ops/us
PoolBenchmark.acquireAndRelease:released       32         128                    16       0               false  thrpt   10  6.713 ± 0.057  ops/us
PoolBenchmark.acquireAndRelease                32         128                    16       0                true  thrpt   10  0.220 ± 0.001  ops/us
PoolBenchmark.acquireAndRelease:acquired       32         128                    16       0                true  thrpt   10  7.032 ± 0.045  ops/us
PoolBenchmark.acquireAndRelease:released       32         128                    16       0                true  thrpt   10  7.032 ± 0.045  ops/us
``` 
Which shows that the workstealing operation is not the main bottleneck here (the counter update, instead, is :"( ) given that the performance hit is not really huge.

For the workstealing use case instead, performance are pretty much on par of the existing pool, but adding GC profiling shows an interesting picture here:
```
Benchmark                                           (burst)  (capacity)  (fakeEventExecutors)  (work)  (workStealingPool)   Mode  Cnt   Score    Error   Units
PoolBenchmark.acquireAndRelease                          32         128                     4       0               false  thrpt   10   0.213 ±  0.009  ops/us
PoolBenchmark.acquireAndRelease:acquired                 32         128                     4       0               false  thrpt   10   6.822 ±  0.273  ops/us
PoolBenchmark.acquireAndRelease:gc.alloc.rate            32         128                     4       0               false  thrpt   10   5.113 ±  0.350  MB/sec
PoolBenchmark.acquireAndRelease:gc.alloc.rate.norm       32         128                     4       0               false  thrpt   10  25.202 ±  2.729    B/op
PoolBenchmark.acquireAndRelease:gc.count                 32         128                     4       0               false  thrpt   10     ≈ 0           counts
PoolBenchmark.acquireAndRelease:released                 32         128                     4       0               false  thrpt   10   6.822 ±  0.273  ops/us
PoolBenchmark.acquireAndRelease                          32         128                     4       0                true  thrpt   10   0.216 ±  0.022  ops/us
PoolBenchmark.acquireAndRelease:acquired                 32         128                     4       0                true  thrpt   10   6.922 ±  0.719  ops/us
PoolBenchmark.acquireAndRelease:gc.alloc.rate            32         128                     4       0                true  thrpt   10   0.009 ±  0.001  MB/sec
PoolBenchmark.acquireAndRelease:gc.alloc.rate.norm       32         128                     4       0                true  thrpt   10   0.046 ±  0.005    B/op
PoolBenchmark.acquireAndRelease:gc.count                 32         128                     4       0                true  thrpt   10     ≈ 0           counts
PoolBenchmark.acquireAndRelease:released                 32         128                     4       0                true  thrpt   10   6.922 ±  0.719  ops/us
```
which shows that the new pool doesn't allocate in the hot path, which is a plus. 
The reason is that the reentrant lock(s) in the queue requires allocating a wait list under contention, while the new pool doesn't need such.

franz1981:
And fec94bd7522f2fe2969041ae1ff4d5ff45fd84b7 shows that I didn't protected the counters good enough from false-sharing: adding a more paranoid padding (https://stackoverflow.com/questions/29199779/false-sharing-and-128-byte-alignment-padding for some info why) increase a performance in pretty much all cases (yay).

franz1981:
Revised numbers including the new padding scheme, for 

2 Threads:
```
Benchmark                                 (burst)  (capacity)  (fakeEventExecutors)  (work)  (workStealingPool)   Mode  Cnt   Score   Error   Units
PoolBenchmark.acquireAndRelease                 1        1024                     0       0               false  thrpt   10   2.584 ± 0.242  ops/us
PoolBenchmark.acquireAndRelease:acquired        1        1024                     0       0               false  thrpt   10   2.584 ± 0.242  ops/us
PoolBenchmark.acquireAndRelease:released        1        1024                     0       0               false  thrpt   10   2.584 ± 0.242  ops/us
PoolBenchmark.acquireAndRelease                 1        1024                     0       0                true  thrpt   10  13.225 ± 0.549  ops/us
PoolBenchmark.acquireAndRelease:acquired        1        1024                     0       0                true  thrpt   10  13.225 ± 0.549  ops/us
PoolBenchmark.acquireAndRelease:released        1        1024                     0       0                true  thrpt   10  13.225 ± 0.549  ops/us
```
4 Threads:
```
Benchmark                                 (burst)  (capacity)  (fakeEventExecutors)  (work)  (workStealingPool)   Mode  Cnt   Score   Error   Units
PoolBenchmark.acquireAndRelease                 1        1024                     0       0               false  thrpt   10   6.427 ± 0.050  ops/us
PoolBenchmark.acquireAndRelease:acquired        1        1024                     0       0               false  thrpt   10   6.427 ± 0.050  ops/us
PoolBenchmark.acquireAndRelease:released        1        1024                     0       0               false  thrpt   10   6.427 ± 0.050  ops/us
PoolBenchmark.acquireAndRelease                 1        1024                     0       0                true  thrpt   10  11.781 ± 0.797  ops/us
PoolBenchmark.acquireAndRelease:acquired        1        1024                     0       0                true  thrpt   10  11.782 ± 0.797  ops/us
PoolBenchmark.acquireAndRelease:released        1        1024                     0       0                true  thrpt   10  11.782 ± 0.797  ops/us
```
and, with workstealing

i.e. 
```
$ java -jar target/benchmarks.jar PoolBenchmark -pworkStealingPool=false,true -pcapacity=128 -pburst=32 -t 4 -pfakeEventExecutors=4

Benchmark                                 (burst)  (capacity)  (fakeEventExecutors)  (work)  (workStealingPool)   Mode  Cnt  Score   Error   Units
PoolBenchmark.acquireAndRelease                32         128                     4       0               false  thrpt   10  0.196 ± 0.006  ops/us
PoolBenchmark.acquireAndRelease:acquired       32         128                     4       0               false  thrpt   10  6.271 ± 0.186  ops/us
PoolBenchmark.acquireAndRelease:released       32         128                     4       0               false  thrpt   10  6.271 ± 0.186  ops/us
PoolBenchmark.acquireAndRelease                32         128                     4       0                true  thrpt   10  0.243 ± 0.008  ops/us
PoolBenchmark.acquireAndRelease:acquired       32         128                     4       0                true  thrpt   10  7.787 ± 0.259  ops/us
PoolBenchmark.acquireAndRelease:released       32         128                     4       0                true  thrpt   10  7.787 ± 0.259  ops/us
```
while removing the fake executors (while keeping the burst in order to have a fair comparison)
```
Benchmark                                 (burst)  (capacity)  (fakeEventExecutors)  (work)  (workStealingPool)   Mode  Cnt   Score   Error   Units
PoolBenchmark.acquireAndRelease                32         128                     0       0               false  thrpt   10   0.210 ± 0.009  ops/us
PoolBenchmark.acquireAndRelease:acquired       32         128                     0       0               false  thrpt   10   6.733 ± 0.286  ops/us
PoolBenchmark.acquireAndRelease:released       32         128                     0       0               false  thrpt   10   6.733 ± 0.286  ops/us
PoolBenchmark.acquireAndRelease                32         128                     0       0                true  thrpt   10   0.345 ± 0.029  ops/us
PoolBenchmark.acquireAndRelease:acquired       32         128                     0       0                true  thrpt   10  11.053 ± 0.924  ops/us
PoolBenchmark.acquireAndRelease:released       32         128                     0       0                true  thrpt   10  11.053 ± 0.924  ops/us
```
it shows that workstealing still make the new pool to have an edge vs the old one, but NOT having workstealing get the best results - pretty much always doubling performance compared to the original pool.

Right now the sole concern is about `reserve` cost, and I still need to publish numbers about it.
```

**Comment by franz1981** [comment]
After the changes https://github.com/Hyperfoil/Hyperfoil/pull/374/commits/b3b6b1a99d2cb7d63e9d246ab554097f8e63cb24
we can now observe what's going on for real (without class loading in) in the `SingleShotTime` benchmark:
```
Benchmark                     (capacity)  (eventExecutors)  Mode  Cnt      Score   Error  Units
PoolReserveBenchmark.reserve         128                 0    ss    2     69.240          us/op
PoolReserveBenchmark.reserve         128                 1    ss    2    127.504          us/op
PoolReserveBenchmark.reserve         128                 2    ss    2    118.445          us/op
PoolReserveBenchmark.reserve         128                 4    ss    2    135.010          us/op
PoolReserveBenchmark.reserve        1024                 0    ss    2    360.428          us/op
PoolReserveBenchmark.reserve        1024                 1    ss    2    512.362          us/op
PoolReserveBenchmark.reserve        1024                 2    ss    2    513.437          us/op
PoolReserveBenchmark.reserve        1024                 4    ss    2    502.958          us/op
PoolReserveBenchmark.reserve      131072                 0    ss    2   6172.179          us/op
PoolReserveBenchmark.reserve      131072                 1    ss    2   9973.640          us/op
PoolReserveBenchmark.reserve      131072                 2    ss    2  10117.694          us/op
PoolReserveBenchmark.reserve      131072                 4    ss    2  10098.414          us/op
```
There's clearly a performance impact but is not as cathastrofic as it was before the changes

franz1981:
This change could have bee way better in 2 ways:
1. by ensuring that no alien threads can use the new pool: this is sadly not true on https://github.com/Hyperfoil/Hyperfoil/blob/ac5c4c1bb1713e9bcf8c99f3351ba7e6269b8541/core/src/main/java/io/hyperfoil/core/impl/SimulationRunner.java#L367 which can start from a non event loop thread -> and we should evaluate why and if it can be fixed, too
2. "precise" stats tracking (for min/max) is causing a cache traffic which doesn't scale (even with `LongAdder` ) because for a partitioned counters (which has been measured with the `CountersScalability` benchmark) tracking min/max still requires to query to have the other partitions current value, causing an invalidation of the cache line where the partitioned counter belong, which both slow down reader and writer. 

The last point can be summarized, using the simplified `MESI` protocol like this:
- a counter is updated to the local cache, becoming `M` (Modified) - or `E` (Exclusive)
- a reader thread already have a version of the same counter, which is `I` (Invalid) given that has been updated by the writer
- the same reader thread query for a fresh new value, moving it to be `S` (Shared) for both itself and the writer cache
- while becoming `S` for both, if its value was different from the memory one, is updated on the memory too
- the writer thread modify it again twice, making it `E` (Exclusive) and `S`, notifying at the first transition to `E` the copy in the reader cache to be `I` (Invalid)

In short, the fact that a reader keep on asking for up to date values, slow down both, causing cache coherency traffic too.

lampajr (approved):
Thanks @franz1981 for all the details, I still need to get used to the micro-benchmarking and you already put some real good examples here!!!

I just left a couple of minors but it is ready to go IMO! :heart:

benchmarks/src/main/java/io/hyperfoil/core/harness/FastThreadLocalThreadHarnessExecutor.java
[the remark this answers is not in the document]
franz1981: Effectively is not required anymore - so I can drop it - good catch!

lampajr (approved):
Awesome!

Feel free to merge it @franz1981 :rocket:

Issue #361, opened 2024-06-10
Opt-out HTTP caching


franz1981:
Http caching is very memory intensive (both due to `CacheControl`(s) and `HttpCacheImpl` instances): the original `wrk` and `wrk2` tools doesn't support caching, hence, for high throughput scenario (which would require a pretty big amount of `Session`s) would be great to disable client-side caching at all.

This would lower the barrier of getting a simpler experience while tuning the client machine running Hyperfoil.

PR #375, opened 2024-06-26
Make HTTP cache configurable

lampajr:
Note that this is based on https://github.com/Hyperfoil/Hyperfoil/pull/372, therefore this might be merged only once the other one is merged and this one properly rebased.
 
## Fixes Issue

Fixes https://github.com/Hyperfoil/Hyperfoil/issues/361

## Changes proposed

Made the HTTP cache configurable as there might be use cases, and then benchmarks, that do not require any HTTP cache.

By default the HTTP cache is enabled, mainly for backward compatibility reasons but it can be disabled by adding `useHttpCache: false` in the `http` definition:
```yaml
name: first-benchmark
http:
  host: http://localhost:8080
  sharedConnections: 10
  useHttpCache: false
```
> [!NOTE]
> If there are multiple `http` authorities, you need to disable the cache on all of them to really disable it.

Disabling the HTTP cache (if not really required) will heavily reduce the amount allocated java object (and therefore the used heap) during the benchmark simulation initialization.

TODOs:
- [ ] Planning to provide some before/after comparisons in the comments.
- [x] Update the documentation including the new `useHttpCache` option (edit: https://github.com/Hyperfoil/hyperfoil.github.io/pull/60).

<details>
<summary>
</summary>

</details>

### DIFF
--- a/cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
+++ b/cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
@@ -155,6 +155,9 @@ public abstract class WrkAbstract {
       @Option(name = "enable-http2", description = "HTTP2 is not supported in wrk/wrk2: you can enable that for Hyperfoil.", defaultValue = "false")
       boolean enableHttp2;
 
+      @Option(name = "use-http-cache", description = "By default the HTTP cache is disabled, providing this option you can enable it.", hasValue = false)
+      boolean useHttpCache;
+
       @Argument(description = "URL that should be accessed", required = true)
       String url;
 
@@ -220,6 +223,7 @@ public abstract class WrkAbstract {
                      .protocol(protocol).host(uri.getHost()).port(protocol.portOrDefault(uri.getPort()))
                      .allowHttp2(enableHttp2)
                      .sharedConnections(connections)
+                     .useHttpCache(useHttpCache)
                   .endHttp()
                .endPlugin()
                .threads(this.threads);
--- a/http/src/main/java/io/hyperfoil/http/api/HttpRequest.java
+++ b/http/src/main/java/io/hyperfoil/http/api/HttpRequest.java
@@ -24,11 +24,12 @@ public class HttpRequest extends Request {
    public HttpMethod method;
    public String authority;
    public String path;
-   public CacheControl cacheControl = new CacheControl();
+   public final CacheControl cacheControl;
    private HttpConnectionPool pool;
 
-   public HttpRequest(Session session) {
+   public HttpRequest(Session session, boolean httpCacheEnabled) {
       super(session);
+      this.cacheControl = httpCacheEnabled ? new CacheControl() : null;
    }
 
    public static HttpRequest ensure(Request request) {
@@ -56,7 +57,7 @@ public class HttpRequest extends Request {
          // Handlers must not block anyway, so this is illegal way to run request
          // and happens only with programmatic configuration in testsuite.
          throw new IllegalStateException(String.format(
-               "#{} Invoking request directly from a request handler; current: {}, requested {}",
+               "#%d Invoking request directly from a request handler; current: %s, requested %s",
                session.uniqueId(), session.currentRequest(), this));
       }
 
@@ -78,7 +79,9 @@ public class HttpRequest extends Request {
       this.authority = null;
       this.path = null;
       this.pool = null;
-      cacheControl.reset();
+      if (this.cacheControl != null) {
+         this.cacheControl.reset();
+      }
    }
 
    public HttpResponseHandlers handlers() {
@@ -128,4 +131,16 @@ public class HttpRequest extends Request {
          session.proceed();
       }
    }
+
+   /**
+    * Checks if the cache control is set.
+    * Worth to note that, if this is true the tool guarantees that HttpCache.get(request.session) is not null.
+    *
+    * @return {@code true} if {@code cacheControl} is not {@code null},
+    *         indicating that HTTP cache is enabled; {@code false} otherwise.
+    */
+   public boolean hasCacheControl() {
+      return this.cacheControl != null;
+   }
+
 }
--- a/http/src/main/java/io/hyperfoil/http/connection/Http1xConnection.java
+++ b/http/src/main/java/io/hyperfoil/http/connection/Http1xConnection.java
@@ -5,17 +5,20 @@ import java.util.Deque;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
 import io.hyperfoil.api.connection.Connection;
-import io.hyperfoil.http.api.HttpVersion;
 import io.hyperfoil.api.session.Session;
 import io.hyperfoil.api.session.SessionStopException;
-import io.hyperfoil.impl.Util;
 import io.hyperfoil.http.api.HttpCache;
 import io.hyperfoil.http.api.HttpConnection;
 import io.hyperfoil.http.api.HttpConnectionPool;
 import io.hyperfoil.http.api.HttpRequest;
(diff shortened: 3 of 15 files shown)


lampajr:
>Note that this is based on https://github.com/Hyperfoil/Hyperfoil/pull/372, therefore this might be merged only once the other one is merged and this one properly rebased.

Properly rebased on top of `master` branch!

franz1981:
Perfect, well done! I will give another round of review today and we're ready to go!

cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
```diff
@@ -220,6 +220,7 @@ public CommandResult execute(HyperfoilCommandInvocation invocation) {
                      .protocol(protocol).host(uri.getHost()).port(protocol.portOrDefault(uri.getPort()))
                      .allowHttp2(enableHttp2)
                      .sharedConnections(connections)
+                     .useHttpCache(false)
```
franz1981, line 223: We can still have it opt-in adding a specific wrk argument to enable it back, wdyt?
this will help you to check with/without and users as well

franz1981 (changes requested):

http/src/main/java/io/hyperfoil/http/connection/Http1xConnection.java
```diff
@@ -159,8 +159,12 @@ public void request(HttpRequest request,
          buf.writeByte('\r').writeByte('\n');
       }
 
-      HttpCache httpCache = HttpCache.get(request.session);
-      httpCache.beforeRequestHeaders(request);
+      HttpCache httpCache = null;
+      if (request.hasCacheControl()) {
```
franz1981, line 163: it is still fine to have this method, I like it - just mask it with a null check

cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
franz1981: yeah, in this case, it could be handy - as long as there are no conflicting named wrk2 args

http/src/main/java/io/hyperfoil/http/api/HttpRequest.java
```diff
@@ -24,11 +24,14 @@ public class HttpRequest extends Request {
    public HttpMethod method;
    public String authority;
    public String path;
-   public CacheControl cacheControl = new CacheControl();
+   private final boolean httpCacheEnabled;
```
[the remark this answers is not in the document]
lampajr: Overall the instance size is 88 bytes, with or without this boolean.. due to some `alignment/padding gap` we reach `88` in both cases (at least based on the JOL).

On the other hand, I agree that there is no particular reason in keeping this boolean as it is just kept to avoid `cacheControl != null`. I am expecting the difference between the two checks is in terms of nanoseconds, so I think it could be considered a negligibile, right?

lampajr:
Thanks @franz1981, I applied all you suggestions and I did very simple test just to provide evidence of the reduced amount of allocation (and therefore used heap) at start up:

Running
```sh
wrk2 -c 10 -t 2 -R 500 http://localhost:8080/time/cached [--use-http-cache]
```

Without cache (1st run, which starts when you see the first heap increase):
```
18:04:21,569 INFO  (vert.x-eventloop-thread-3) [i.h.c.i.SimulationRunner] Running additional GC check!
18:04:21,660 INFO  (vert.x-eventloop-thread-3) [i.h.c.i.SimulationRunner] GC execution took 91 ms
18:04:21,700 INFO  (vert.x-eventloop-thread-3) [i.h.c.i.SimulationRunner] Overall GC run took 131 ms
18:04:21,701 INFO  (vert.x-eventloop-thread-3) [i.h.c.i.SimulationRunner] Simulation initialization took 174 ms
```
With cache (2nd run):
```
18:04:54,262 INFO  (vert.x-eventloop-thread-5) [i.h.c.i.SimulationRunner] Running additional GC check!
18:04:54,390 INFO  (vert.x-eventloop-thread-5) [i.h.c.i.SimulationRunner] GC execution took 128 ms
18:04:54,431 INFO  (vert.x-eventloop-thread-5) [i.h.c.i.SimulationRunner] Overall GC run took 169 ms
18:04:54,431 INFO  (vert.x-eventloop-thread-5) [i.h.c.i.SimulationRunner] Simulation initialization took 233 ms
```
Without cache (3rd run):
```
18:05:17,810 INFO  (vert.x-eventloop-thread-7) [i.h.c.i.SimulationRunner] Running additional GC check!
18:05:17,901 INFO  (vert.x-eventloop-thread-7) [i.h.c.i.SimulationRunner] GC execution took 91 ms
18:05:17,941 INFO  (vert.x-eventloop-thread-7) [i.h.c.i.SimulationRunner] Overall GC run took 131 ms
18:05:17,942 INFO  (vert.x-eventloop-thread-7) [i.h.c.i.SimulationRunner] Simulation initialization took 144 ms
```

image

cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
lampajr: Added as `--use-http-cache`, by default it is set to `false`.

http/src/main/java/io/hyperfoil/http/connection/Http1xConnection.java
```diff
@@ -366,7 +371,9 @@ public void putHeader(CharSequence header, CharSequence value) {
          }
          buf.writeByte('\r');
          buf.writeByte('\n');
-         HttpCache.get(request.session).requestHeader(request, header, value);
+         if (request.hasCacheControl()) {
```
franz1981, line 374: if the httpCache is not null it should always have the cache control (as it was before?)

http/src/main/java/io/hyperfoil/http/steps/ClearHttpCacheAction.java
```diff
@@ -10,7 +10,10 @@
 public class ClearHttpCacheAction implements Action {
    @Override
    public void run(Session session) {
-      HttpCache.get(session).clear();
+      HttpCache httpCache = HttpCache.get(session);
```
franz1981, line 13: Could we drop the `ClearHttpCacheAction` if `httpCache` is not enabled?
This would avoid querying the `httpCache` in the session and having this action in as a whole

http/src/main/java/io/hyperfoil/http/steps/HttpResponseHandlersImpl.java
```diff
@@ -286,7 +290,10 @@ public void handleEnd(HttpRequest request, boolean executed) {
                      handler.after(request.session);
                   }
                }
-               HttpCache.get(request.session).tryStore(request);
+               HttpCache httpCache = HttpCache.get(request.session);
```
franz1981, line 293: it would be great to not query it if we know there should be none (by checking the cache control first?)

franz1981 (changes requested):

http/src/main/java/io/hyperfoil/http/steps/ClearHttpCacheAction.java
lampajr: Mmmh IIRC `ClearHttpCacheAction` is a step that the use should manually set in the benchmark, so if the user does not explicitly set that action it should appear at all at runtime.

http/src/main/java/io/hyperfoil/http/connection/Http1xConnection.java
lampajr: yeah that's true, do you see something wrong here?

http/src/main/java/io/hyperfoil/http/steps/HttpResponseHandlersImpl.java
lampajr: Done :white_check_mark:

franz1981 (approved):
Great job!!! Ready to go IMO!!

Issue #380, opened 2024-07-04
On `wrk` the session allocates `maxRequests`-size HttpRequestPool


lampajr:
On `wrk` the session allocates by default `maxRequests`-size (16) HttpRequestPool even if the default behavior on the original `wrk` is to use 1 single request concurrency per connection.

This could be related to the pipelining (see https://github.com/Hyperfoil/Hyperfoil/issues/277).

lampajr:
Looking at this problem we noticed that we are currently storing the HttpRequestPool on every session https://github.com/Hyperfoil/Hyperfoil/blob/80c2bfc2f483ee2baf53ea26cbbe830305310462/http/src/main/java/io/hyperfoil/http/HttpRunData.java#L116

This could end up in huge allocation when using an high rates.

lampajr:
Generally the `maxRequests` should be set to the `pipelining limit` * `shared connections`.

franz1981:
This is for reference a screenshot of a run with wrk2 with rate 100K rps, using 10 connections:
![image](https://github.com/Hyperfoil/Hyperfoil/assets/13125299/228dfcfe-f1db-41c6-97e4-3fa53c738c60)

franz1981:
FYI @lampajr just setting it == number of connections on WrkAbstract at  https://github.com/Hyperfoil/Hyperfoil/blob/80c2bfc2f483ee2baf53ea26cbbe830305310462/cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java#L330 on the scenario (we have 2 for wrk actually...)

we have

![image](https://github.com/Hyperfoil/Hyperfoil/assets/13125299/ded1bc6b-95da-4390-9362-fad156d7d367)

which is way better but still. IDK how these are populated; they just look to be way too much to me TBH.
I need to debug and see what happen to understand how we create them...

franz1981:
Using the modified version using 10 maxrequests I see something interesting:
![image](https://github.com/Hyperfoil/Hyperfoil/assets/13125299/440abc3a-4c25-47c0-8edc-52876e093477)

we have 4 scenarios created, and still...

franz1981:
And indeed I see that each sessions got its own configured HttpRequestPool with its own HttpRequest (not shared)!

franz1981:
[...]

PR #382, opened 2024-07-06
Reduce maxRequests to 1 for Wrk HTTP 1.1

franz1981:
## Changes proposed

This is just better configuring `maxRequests` while waiting to better implement `HttpRequestPool` to 
fix it by automatically estimate it - making the `maxRequests` configuration option obsolete.

### DIFF
--- a/cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
+++ b/cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
@@ -330,6 +330,8 @@ public abstract class WrkAbstract {
          if (!enableHttp2) {
             // given that we don't support pipelining we can just safely limit the maxRequests to 1
             scenarioBuilder.maxRequests(1);
+            // each session can have only one sequence
+            scenarioBuilder.maxSequences(1);
          }
          return scenarioBuilder
                   .initialSequence("request")
--- a/cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
+++ b/cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
@@ -323,10 +323,15 @@ public abstract class WrkAbstract {
          String[][] parsedHeaders = this.parsedHeaders;
          long duration = Util.parseToMillis(durationStr);
          // @formatter:off
-         return phaseConfig(benchmarkBuilder.addPhase(phase))
+         var scenarioBuilder = phaseConfig(benchmarkBuilder.addPhase(phase))
                  .duration(duration)
                  .maxDuration(duration + Util.parseToMillis(timeout))
-                 .scenario()
+                 .scenario();
+         if (!enableHttp2) {
+            // given that we don't support pipelining we can just safely limit the maxRequests to 1
+            scenarioBuilder.maxRequests(1);
+         }
+         return scenarioBuilder
                   .initialSequence("request")
                      .step(SC).httpRequest(HttpMethod.GET)
                         .path(path)


franz1981:
I don't know if we need any benchmark considering that this reduce the overall memory footprint of http requests from the default 16 per session, to just 1, which is just better.

This is a temporary fix while waiting for https://github.com/Hyperfoil/Hyperfoil/issues/380#issuecomment-2211769519

As another follow up "simple" fix I could already preset a default `maxRequests` to 1 (instead of 16) if Http 2 upgrade is not possible.

franz1981:
@lampajr 

This can be both a good change or something which breaks badly wrk in some unexpected ways: we need to write down the expected lifecycle of a `Session` during the HTTP 1.1 request handling, to better understand the purpose of the http request pool within it and if we can assume that the `Session` cannot be reused (with its pool) till its associated response isn't completed (or failed/timedout) and release back to the pool.
If this last statement is correct, the PR here will work as expected, because is just impossible that a `Session` with an empty pool (which single request has not yet completed) will be (re)used on a  `PrepareHttpRequestStep` to acquire a new request, failing.

franz1981:
I've analyzed better what happen under the hood with both wrk and wrk2 and found that, both uses https://github.com/Hyperfoil/Hyperfoil/blob/master/http/src/main/java/io/hyperfoil/http/steps/HttpRequestStepBuilder.java#L557 and https://github.com/Hyperfoil/Hyperfoil/blob/master/http/src/main/java/io/hyperfoil/http/steps/HttpRequestStepBuilder.java#L577 to prevent the current (user) `Session` to complete till the response is not fully received and notify https://github.com/Hyperfoil/Hyperfoil/blob/master/http/src/main/java/io/hyperfoil/http/steps/HttpRequestStepBuilder.java#L559 which unblock `AfterSyncRequestStep`.

This configuration prevent a `Session` which `request` has already been used to release it back to its pool till `AfterSyncRequestStep` return true.
And by consequence it means that a single `Session` won't be made available to run any new request till its previous one is not fully completed, making the change in this PR safe.

Instead, to further add details of what happen under the hood:
- `Wrk` allocate an [`Always`](https://github.com/Hyperfoil/Hyperfoil/blob/3417575207e5c8d14f83a59213ceb87c8fcfc216/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java#L318-L358) phase instance using `users === connections` which re-start a finished `Session` (to re-run the phase) when `AfterSyncRequestStep` make it possible: it means that the max required number of sessions is the same of configured connections
- `Wrk2` instead uses [`OpenModelPhase`](https://github.com/Hyperfoil/Hyperfoil/blob/3417575207e5c8d14f83a59213ceb87c8fcfc216/core/src/main/java/io/hyperfoil/core/impl/OpenModelPhase.java) which return back the `Session` acquired on `proceed` to the session pool, while finished: it means that `AfterSyncRequestStep` always save an existing request pool to be used by another request/phase execution, but the number of required `Session` depends by how fast the requests complete vs the configured rate. In the worst ever case we would need a new `Session` for each expected fire time/request for the whole duration of the test

franz1981:
These are the synthetic improvements by running wrk2 with 100 connections, 2 threads and 100 K rate:

before this PR

image

this PR

image

Which shows that now we can (this could be improved a lot yet!) work with 6.8 GB heap while before this change we needed 13.8 GB. 
The startup as well has improved too due to the reduced number of allocations.

franz1981:
@lampajr what popup to my eyes here is that we should likely run another GC after the calibration phase - or better, before a new phase start, too.

In the `this PR` figure you can see that after the  `calibration` phase, the newly allocated sessions/requests etc etc are piling up to the existing garbage i.e. the previous pooled data, because my machine has 64 GB of memory and by default the JVM uses 1/4 of it, if available i.e. 16 GB. Due to this we risk to have a much longer phase while starting the new phase because it has to deal with the previous GC (which can be collected now, as the rest of the figure show).

franz1981:
I've added an additional commit doing something similar for `SequenceInstance` and getting

image

image

lampajr (approved):
Thanks @franz1981, LGTM!!

This is a great example of small change that have huge improvements!

> This configuration prevent a Session which request has already been used to release it back to its pool till AfterSyncRequestStep return true.
And by consequence it means that a single Session won't be made available to run any new request till its previous one is not fully completed, making the change in this PR safe.

That's really a good finding!!!

>Wrk allocate an Always phase instance using users === connections which re-start a finished Session (to re-run the phase) when AfterSyncRequestStep make it possible: it means that the max required number of sessions is the same of configured connections

That's true because we have HTTP 1.x without pipelining, right? Therefore, the dominating factor is the number of connections, thus having more sessions than connections will just allocate sessions that won't be used. Am i correct?

>what popup to my eyes here is that we should likely run another GC after the calibration phase - or better, before a new phase start, too.

That's interesting, given that at the moment we are not executing it.. what kind of issue we might face? that at some time we can see some GC pauses that might create some pause during the phase execution?

franz1981:
Another thing which is interesting to investigate (I have to fill an issue for this), still related pooling, is the impact of the session pool policy when we have so many sessions:
- FIFO (as it is now) is easy to implement scalable concurrency, but force using all the sessions allocated, and eventually bringing from the main memory each of them, over and over based on the duration of the benchmark 
- LIFO; very complex to implement it in a scalable way, but, if the required sessions are not **all** pooled ones, will be way more cache friendly 

Something to prove with some cache misses analysis :P @lampajr 

> That's true because we have HTTP 1.x without pipelining, right? 

I would say yes, but I have yet to see how a non wrk benchmark with pipelining behave, because there is an opened issue with Http 1.1 and pipelining which make me think there is something broken there...

franz1981:
So, ready to go, I'm merging this :+1: thanks @lampajr !

PR #387, opened 2024-07-10
Fixes #385 Optimize 0 and 1 Sequence's concurrency on HTTP

### LINKED ISSUE #385: Investigate BitSetResource 0 and 1 special cases
Looking a the heap dump for wrk/wrk2 most of the BitSetResource used on `BeforeSyncRequestStep` have 0 concurrency.
Both 0 and 1 cases can be treated with some simpler data structure or just in a different way to further reduce the impact on `Session`s resource footprint, given that `BeforeSyncRequestStep` is reserving a new  `BitSetResource` for each `Session` allocated (which is bad for wrk2 with high rate, in particular).

franz1981:
## Fixes Issue #385 

- [x] Added tests stressing the optimized code path

Thanks to this change we're shaving another GB out of 4 (i.e. using than ~3 GB) while using `wrk2 -c 100 -r 100000` .

This improvement should hold outside wrk scripts, depending how many sessions are allocated by a benchmark.

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/util/BitSetResource.java
+++ b/core/src/main/java/io/hyperfoil/core/util/BitSetResource.java
@@ -1,11 +1,86 @@
 package io.hyperfoil.core.util;
 
-import java.util.BitSet;
-
 import io.hyperfoil.api.session.Session;
 
-public class BitSetResource extends BitSet implements Session.Resource {
-   public BitSetResource(int nbits) {
-      super(nbits);
+import java.util.BitSet;
+
+public interface BitSetResource extends Session.Resource {
+
+   void set(int index);
+
+   boolean get(int index);
+
+   void clear(int index);
+
+   static BitSetResource with(int nbits) {
+      if (nbits == 0 || nbits == 1) {
+         class SingleBitSetResource implements BitSetResource {
+
+            private boolean set;
+
+            @Override
+            public void set(int index) {
+               validateBitIndex(index);
+               set = true;
+            }
+
+            private static void validateBitIndex(int index) {
+               if (index != 0) {
+                  throw new IndexOutOfBoundsException();
+               }
+            }
+
+            @Override
+            public boolean get(int index) {
+               validateBitIndex(index);
+               return set;
+            }
+
+            @Override
+            public void clear(int index) {
+               validateBitIndex(index);
+               set = false;
+            }
+         }
+
+         return new SingleBitSetResource();
+      }
+      class MultiBitSetResource extends BitSet implements BitSetResource {
+
+         private final int nBits;
+
+         MultiBitSetResource(int nbits) {
+            super(nbits);
+            this.nBits = nbits;
+         }
+
+         @Override
+         public void set(final int bitIndex) {
+            validateBitIndex(bitIndex);
+            super.set(bitIndex);
+         }
+
+         private void validateBitIndex(int bitIndex) {
+            if (bitIndex >= nBits) {
+               throw new IndexOutOfBoundsException();
+            }
+         }
+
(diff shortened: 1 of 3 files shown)


franz1981:
@lampajr FYI 

https://github.com/Hyperfoil/Hyperfoil/pull/387/files#diff-1bf8b93bca2dc2dc7679e2e3880c22e058990f53a2b9e8ec01d626c2c373d779R16

is due to

https://github.com/Hyperfoil/Hyperfoil/blob/f885994810083f295ab0841b0fb2c059582523a6/api/src/main/java/io/hyperfoil/api/config/SequenceBuilder.java#L33-L35

and the only tests which was using this "feature" was https://github.com/Hyperfoil/Hyperfoil/blob/715fbdb9ab1014b3290b77883a6207f07c364d2d/http/src/test/java/io/hyperfoil/http/html/EmbeddedResourcesTest.java

I could have optimize it in order to have a long field - able to deal with 0->63 different values for concurrency, without using `BitSet` but I still preferred to have this simplified version instead.

core/src/main/java/io/hyperfoil/core/util/BitSetResource.java
```diff
@@ -1,11 +1,88 @@
...
+
+    void set(int index);
+
+    boolean get(int index);
+
+    void clear(int index);
+
+    static BitSetResource with(int nbits) {
+        if (nbits == 0 || nbits == 1) {
+            class SingleBitSetResource implements BitSetResource {
```
lampajr, line 17: Can't we move `SingleBitSetResource` and `MultiBitSetResource` definition outside the `if-else` (either as static classes or even in a sub-package)? I am not a fan of class definition inside the flow but maybe there is a valid reason.

core/src/main/java/io/hyperfoil/core/util/BitSetResource.java
lampajr, line 1: Looks like the formatting is not properly set, do you mind to reformat it usinghttps://github.com/Hyperfoil/Hyperfoil/blob/master/config/codestyle.xml (maybe you already have it in IDEA)

lampajr (commented):
Looks awesome, I left just a couple of comments mainly from readability perspective.

core/src/main/java/io/hyperfoil/core/util/BitSetResource.java
franz1981: The main reason was to avoid having 2 separate small classes just for few lines of codes.
Sadly java interfaces doesn't allow to define private classes within them, apart with this "trick" I have used (the anonymous/hidden definitions are still just 2)

franz1981: Said that I am not biased into doing like this at any cost, and instead I'm opened to move them into their separate files too

core/src/main/java/io/hyperfoil/core/util/BitSetResource.java
franz1981: I was expecting the compilation to outline this...why is not happening?

lampajr: yeah I was thinking the same but it looks like it is not automated/integrated in the maven lifecycle :face_with_diagonal_mouth:

core/src/main/java/io/hyperfoil/core/util/BitSetResource.java
lampajr: No I think that's fine as it is, I was just curious and to be honest I missed that the outer class is an interface instead :smile:

lampajr (approved):
lgtm!

PR #388, opened 2024-07-10
Extends max sequences and requests wrk settings to HTTP 2 as well

franz1981:
This is a follow up for #382 but applying it to HTTP 2, now that I got better insights about the meaning of such settings.


lampajr (approved):
Awesome, LGTM!

Issue #389, opened 2024-07-11
Wrk/Wrk2 retained dead objects across phases


franz1981:
Performing an heap dump of wrk2 while in the `test` phase reveal that "calibaration" allocated `Sessions` and objects are still retained in the heap, bloating the old generation.
It  would be great to further reduce the `wrk/wrk2` memory requirement by "fixing" this, cutting it in half (or more if we decide to save allocating 15X throughput sessions too!)

lampajr:
Hi @franz1981 I tried as well and I can confirm your finding!

First of all given that we have 2 phases (**calibration** and **test**) the overall number of session is doubled, e.g., with `-c 10 -R 1000` and given that the `maxSession` is set to `15 * rate` we end up with `15k` session per phase (i.e., `30k`).

Just as an example I tried to merge the 2 phase into a single phase with 2 `initialSequences` and it shows already a hug improvement in terms of allocation _but_ it reduce its usability as the user now cannot identify calibration and test results.

So I think that keeping 2 separate phase is a must have unless we see alternatives that are feature-equivalent from the user persepctive.

Thus, I see two options here:
- Reduce the number of `maxSessions`, as you are suggesting and I think this would be mostly a heurstic that we should determine 
- Manually perform the GC after the calibration phase such that we can try to cleanup old objects (here we should check if some objects are there because still referenced somewhere, as we were discussing yesterday)

franz1981:
Thanks for checking @lampajr !
I would rather think on a way to make a single phase of it, which is ignoring statistics coming from the first X seconds - if required, or explore native compilation for Hyperfoil.
Both would address the same problem i.e. warming up the JIT to avoid weird pauses and slowdown due to JIT cost or just cold running code.

lampajr:
> Thanks for checking @lampajr !
> I would rather think on a way to make a single phase of it, which is ignoring or marking differently the statistics coming from the first X seconds or explore native compilation for Hyperfoil.
> Both would address the same problem i.e. warming up the JIT to avoid weird pauses and slowdown due to JIT cost or just cold running code.
> 
> Given that we are mostly GC freeish in the hot path, we do not fear the GC activity - and having something which is immediately fast is appealing, and would save users to guess why they setup wrk to run for 10 seconds and getting 16 instead.

I definitely agree with you, I think that spending some time investigating the native compilation for Hyperfoil is totally worth!

lampajr:
[...]

PR #391, opened 2024-07-12
Better tune calibration wrk2 phase by lowering maxSessions

franz1981:
This is just better tuning the calibration phase of Wrk2 to not use that many sessions, given that they won't be reused
for the test phase, and will still be resident in the old gen of the heap :"(

Simple workaround while waiting #389 to be addressed.

I've adopted the safest possible value here, to avoid missing sessions.

### DIFF
--- a/cli/src/main/java/io/hyperfoil/cli/commands/Wrk.java
+++ b/cli/src/main/java/io/hyperfoil/cli/commands/Wrk.java
@@ -43,7 +43,8 @@ public class Wrk extends WrkAbstract {
    public class WrkCommand extends WrkAbstract.AbstractWrkCommand {
 
       @Override
-      protected PhaseBuilder<?> phaseConfig(PhaseBuilder.Catalog catalog) {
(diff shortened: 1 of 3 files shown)


franz1981:
I could add a test which verify that we cannot requires more sessions in the calibration pahse, too @lampajr wdyt?

lampajr (approved):
Great workaround!!

> I could add a test which verify that we cannot requires more sessions in the calibration pahse, too @lampajr wdyt?

Well, since this is actually a (hopefully) temporary workaround, I don't think it is strictly necessary to have that test. 
On the other hand if adding the test is quite easy, +1 let's add it, otherwise that's fine as well! Up to you :smile:

franz1981:
@lampajr Is up to you to accept this as it is or just wait till you find a better option as in #389

lampajr:
I think we go with this as it is, then if we find better solution as part of the issue you mentioned we can obviously re-adapt the code portion changed here.

PR #400, opened 2024-07-19
Skip benchmark validation run on wrk/wrk2

lampajr:
Partially related to https://github.com/Hyperfoil/Hyperfoil/issues/389.

In the current implemention of `wrk` and `wrk2` wrappers, the command executed `2` different runs:
- #1: This is just a validation run, no actual simulation is performed https://github.com/Hyperfoil/Hyperfoil/blob/52df8711e34d214f6b34e6b1cb45419c158bb61c/clustering/src/main/java/io/hyperfoil/clustering/ControllerVerticle.java#L160-L168
- #2: The second one is the real simulation run

This results in two runs stored:
Screenshot from 2024-07-19 11-24-39

But, most importantly, the 2 runs are separated runs.. this means that the allocation is ~twice the expected one as we are creating two identical runs.

Given that the `wrk` benchmarks are pre-defined benchmarks, I don't see why we should run a validation-only run before running the simulation, but maybe I am missing something here.

## Changes proposed

Assuming my understanding is correct, I am proposing to remove the validation run which should reduce the used heap as only one run will be created.

<details>
<summary>
</summary>

</details>

### DIFF
--- a/cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
+++ b/cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
@@ -163,6 +163,8 @@ public abstract class WrkAbstract {
 
       String path;
       String[][] parsedHeaders;
+      boolean started = false;
+      boolean initialized = false;
 
       @Override
       public CommandResult execute(HyperfoilCommandInvocation invocation) {
@@ -246,7 +248,9 @@ public abstract class WrkAbstract {
          addPhase(builder, PhaseType.calibration, "6s");
          // We can start only after calibration has full completed because otherwise some sessions
          // would not have connection available from the beginning.
-         addPhase(builder, PhaseType.test, duration).startAfterStrict(PhaseType.calibration.name()).maxDuration(Util.parseToMillis(duration));
+         addPhase(builder, PhaseType.test, duration)
+               .startAfterStrict(PhaseType.calibration.name())
+               .maxDuration(Util.parseToMillis(duration));
 
          RestClient client = invocation.context().client();
          if (client == null) {
@@ -258,30 +262,18 @@ public abstract class WrkAbstract {
          Client.BenchmarkRef benchmark = client.register(builder.build(), null);
          invocation.context().setServerBenchmark(benchmark);
 
-         //validate benchmark
-         Client.RunRef run = benchmark.start(null, Collections.emptyMap(), Boolean.TRUE);
+         Client.RunRef run = benchmark.start(null, Collections.emptyMap());
+         invocation.context().setServerRun(run);
 
          boolean result = awaitBenchmarkResult(run, invocation);
 
          if (result) {
-            if (run.get().errors.size() > 0) {
-               invocation.println("ERROR: " + run.get().errors.stream().collect(Collectors.joining(", ")));
+            if (!started && !run.get().errors.isEmpty()) {
+               // here the benchmark simulation did not start, it failed during initialization
+               // print the error that is returned by the controller, e.g., cannot connect
+               invocation.println("ERROR: " + String.join(", ", run.get().errors));
                return CommandResult.FAILURE;
             }
-         } else {
-            return CommandResult.FAILURE;
-
-         }
-
-         run = benchmark.start(null, Collections.emptyMap());
-         invocation.context().setServerRun(run);
-
-         invocation.println("Running " + duration + " test @ " + url);
-         invocation.println("  " + threads + " threads and " + connections + " connections");
-
-         result = awaitBenchmarkResult(run, invocation);
-
-         if (result) {
             RequestStatisticsResponse total = run.statsTotal();
             RequestStats testStats = total.statistics.stream().filter(rs -> PhaseType.test.name().equals(rs.phase))
                   .findFirst().orElseThrow(() -> new IllegalStateException("Error running command: Missing Statistics"));
@@ -292,7 +284,6 @@ public abstract class WrkAbstract {
          } else {
             return CommandResult.FAILURE;
          }
-
       }
 
       private boolean awaitBenchmarkResult(Client.RunRef run, HyperfoilCommandInvocation invocation) {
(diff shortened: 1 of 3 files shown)


lampajr:
Running `wrk2 -R 10000 -c 1000 http://localhost:8080/time/cached` against https://github.com/franz1981/quarkus-profiling-workshop.

Before:
Screenshot from 2024-07-19 11-12-28

After:
Screenshot from 2024-07-19 11-14-01

johnaohara:
The reason for the 2 runs is that Hyperfoil **itself** needs to warm up, so the client side needs to be JIT'd, otherwise it will be running in interpreted mode until it is warmed.  The way(s) to mitigate this would be to;
- build a native client binary 
- share an aot library with pre-compiled code

the wrk2 emulation **might** not be much of a problem, i.e. the load rate **might** be low enough for interpreted code to not cause a recording issue, but the wrk emulation **will** be an issue without warmup/aot/native build

johnaohara:
'Validation' might not have been the best adjective to use in the original PR

lampajr:
Hi @johnaohara!

> The reason for the 2 runs is that Hyperfoil itself needs to warm up, so the client side needs to be JIT'd, otherwise it will be running in interpreted mode until it is warmed. The way(s) to mitigate this would be to;

Isn't this related to the "calibration" phase? (which is different from the "validation" run I am talking here).
I don't see how the "validation" run can warmup the Hyperfoil, given that it simply creates a run without starting the simulation.. it looks like it is just pure validation of the benchmark.

Am I missing something?

johnaohara:
Ah yes, you are correct, my memory is failing me ;)

So 'Validation' is to validate the benchmark can run, capture errors and report errors to user, otherwise the wrk output was a stack trace that is meaningless to the user.

With this PR, what happens if you try to call against and endpoint that is not listening/non existant?

lampajr:
> With this PR, what happens if you try to call against and endpoint that is not listening/non existant?

Ok that's why it was added :smile: 

With validation:
```
./distribution/target/distribution/bin/wrk2.sh -R 10000 -c 1000 http://localhost:8081/wrong                  [12:33:24]
ERROR: in-vm: Cannot connect to localhost:8081: 0 created, 103 failures.
```

Without validation:
```
./distribution/target/distribution/bin/wrk2.sh -R 10000 -c 1000 http://localhost:8081/wrong   [12:32:18]
Running 10s test @ http://localhost:8081/wrong
  2 threads and 1000 connections
Failed to execute command: java.lang.RuntimeException: java.lang.IllegalStateException: Error running command: Missing Statistics
Usage: wrk2 [<options>] <url>
Runs a workload simulation against one endpoint using the same vm

Options:
  -R, --rate         Work rate (throughput)
  -c, --connections  Total number of HTTP connections to keep open
  -d, --duration     Duration of the test, e.g. 2s, 2m, 2h
  -t, --threads      Total number of threads to use.
  -s, --script       !!!NOT SUPPORTED: LuaJIT script
  -h, --help
  -H, --header       HTTP header to add to request, e.g. "User-Agent: wrk"
  --latency          Print detailed latency statistics
  --timeout          Record a timeout if a response is not received within this amount of time.
  -A, --agent        Inline definition of agent executing the test. By default assuming non-clustered mode.
  --enable-http2     HTTP2 is not supported in wrk/wrk2: you can enable that for Hyperfoil.
  --use-http-cache   By default the HTTP cache is disabled, providing this option you can enable it.

Argument:
                   URL that should be accessed
```

johnaohara:
The current behaviour is;

```
$ jbang wrk@Hyperfoil -d 20 -c 10 http://localhost:8090
ERROR: in-vm: Cannot connect to localhost:8090: 0 created, 104 failures.
```
with this PR;

```
$ ./bin/wrk.sh -d 20 -c 10 http://localhost:8090
Running 20 test @ http://localhost:8090
  2 threads and 10 connections
Failed to execute command: java.lang.RuntimeException: java.lang.IllegalStateException: Error running command: Missing Statistics
Usage: wrk [<options>] <url>
Runs a workload simulation against one endpoint using the same vm

Options:
  -c, --connections  Total number of HTTP connections to keep open
  -d, --duration     Duration of the test, e.g. 2s, 2m, 2h
  -t, --threads      Total number of threads to use.
  -s, --script       !!!NOT SUPPORTED: LuaJIT script
  -h, --help
  -H, --header       HTTP header to add to request, e.g. "User-Agent: wrk"
  --latency          Print detailed latency statistics
  --timeout          Record a timeout if a response is not received within this amount of time.
  -A, --agent        Inline definition of agent executing the test. By default assuming non-clustered mode.
  --enable-http2     HTTP2 is not supported in wrk/wrk2: you can enable that for Hyperfoil.
  --use-http-cache   By default the HTTP cache is disabled, providing this option you can enable it.

Argument:
                   URL that should be accessed
```

The output is not clear to the user that the url could not be accessed.   

I am +100 with the intent of this PR, but I think the output should inform the user what the error is

johnaohara:
Our experiments crossed paths ;)

lampajr:
> Our experiments crossed paths ;)

hehe :smile: 

>I am +100 with the intent of this PR, but I think the output should inform the user what the error is

Yeah I agree, thanks a lot for this finding! I completely missed that :pray: 

I will try to see if there are other ways to reach the same goal :crossed_fingers:

franz1981:
Maybe worth using other mechanisms to validate the endpoints instead of running the benchmark...or we could configure it with way less resources, if is a validation run.

johnaohara:
> Maybe worth using other mechanisms to validate the endpoints instead of running the benchmark...or we could configure it with way less resources, if is a validation run.

when i looked at this last year, it required api changes. The changes might be worth it now to save resources

lampajr:
I tried to find _good_ a way to expose a "validation" api but in order to properly validate and check the benchmark, e.g., check whether the provided authority is really reachable, we need to startup the agent (either clustered or not) that will start the "INITIALIZING" phase for the benchmark and only at that point we can validate it - this will require a lot of overhead and unnecessary memory allocation just for the validation.

Therefore, given that this is actually required just for wrk/wrk2, I think the best approach here is to monitor the run `status` from the CLI and:
- If it terminates when still in "INITIALIZING" phase we can safely report the error that was found
- If it terminates when in "RUNNING" phase, let's ignore the errors as the benchmark is valid as it started

```java
            } else if ("INITIALIZING".equals(recent.status) && !initialized) {
               initialized = true;
            } else if ("RUNNING".equals(recent.status) && !started) {
               // here the benchmark simulation started, so we can print wrk/wrk2 messages
               invocation.println("Running " + duration + " test @ " + url);
               invocation.println("  " + threads + " threads and " + connections + " connections");
               started = true;
               // if started, it is also initialized
               // this ensure initialized is set to true if for some reason we did not catch the "INITIALIZING" status
               initialized = true;
            }
```

> So 'Validation' is to validate the benchmark can run, capture errors and report errors to user, otherwise the wrk output was a stack trace that is meaningless to the user.

@johnaohara I think this workaround guarantees that "validation" errors are reported to the users, whereas errors happening during the actual simulation are not reported at all as it was before.

@johnaohara @franz1981 wdyt?

lampajr:
Additional point, if we agree on this approach I can further remove some code that will be no longer required, especially on the CLI side.

franz1981 (approved):
LGTM well done!!!

franz1981:
What's the status of this @lampajr we can close it?

lampajr:
> What's the status of this @lampajr we can close it?

Hey @franz1981, are you talking about https://github.com/Hyperfoil/Hyperfoil/issues/389 ? or this PR specifically? As this one is already merged.

franz1981:
Sorry @lampajr i have likely pasted some wrong comment, can ignore it :P

PR #410, opened 2024-08-01
Migrate to junit 5

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Hyperfoil/issues/373

## Changes proposed

- [x] Migrate the whole test suite to Junit 5
- [x] Refactored some tests to avoid using VertxExtension when not actually used

<details>
<summary>
</summary>

</details>


franz1981:
Thanks for this @lampajr 🙏 

Did you checked https://docs.openrewrite.org/running-recipes/popular-recipe-guides/migrate-from-junit-4-to-junit-5 as well? I suppose it won't work, cause we have vertx in the mix here, but still...

lampajr:
> Thanks for this @lampajr 🙏
> 
> Did you checked https://docs.openrewrite.org/running-recipes/popular-recipe-guides/migrate-from-junit-4-to-junit-5 as well? I suppose it won't work, cause we have vertx in the mix here, but still...

yeah I checked it but I found easier to do it manually as there were many manual changes required due to the `VertxExtension`, moreover going one by one I had change to inspect those tests as well :smile:

lampajr:
Interesting, after rebasing a test started failing in the CI:
```
TwoServersTest.testTwoServers:76->BaseScenarioTest.runScenario:25 » BenchmarkDefinition Phase test, sequence test, step httpRequest (1/3): GET to localhost:0/test is invalid - no HTTP configuration defined.
```

Need to investigate it a bit more!

http/src/test/java/io/hyperfoil/http/TwoServersTest.java
```diff
@@ -37,15 +40,16 @@ protected void initRouter() {
...
+      super.before(vertx, ctx);
       Router secondRouter = Router.router(vertx);
       secondRouter.route("/test").handler(context -> context.response().setStatusCode(300).end());
       secondServer = vertx.createHttpServer().requestHandler(secondRouter)
-            .listen(0, "localhost", ctx.asyncAssertSuccess(srv -> {
+            .listen(0, "localhost", ctx.succeeding(srv -> {
                benchmarkBuilder.plugin(HttpPluginBuilder.class)
                      .http("http://localhost:" + secondServer.actualPort()).endHttp();
+               ctx.completeNow();
             }));
```
lampajr, line 53: > Interesting, after rebasing a test started failing in the CI:
> 
> ```
> TwoServersTest.testTwoServers:76->BaseScenarioTest.runScenario:25 » BenchmarkDefinition Phase test, sequence test, step httpRequest (1/3): GET to localhost:0/test is invalid - no HTTP configuration defined.
> ```
> 
> Need to investigate it a bit more!

The issue seems here, the HTTP plugin is added only once the server is up and running but the test itself does not actually wait for it.. so it might happen that the test starts before the server start listening, thus the issue where Hyperfoil cannot find the http plugin defined in the benchmark.

Was able to simulate locally by adding a `Thread.sleep` just before `benchmarkBuilder.plugin(HttpPluginBuilder.class)`.

franz1981 (approved):
Thanks! well done!

Issue #439, opened 2024-08-30
`TwoServersTest` started failing after #410


lampajr:
**Describe the bug**
After merging https://github.com/Hyperfoil/Hyperfoil/pull/410, master checks started failing at:
```bash
[INFO] Running io.hyperfoil.http.TwoServersTest
Error:  Tests run: 5, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 3.959 s <<< FAILURE! -- in io.hyperfoil.http.TwoServersTest
Error:  io.hyperfoil.http.TwoServersTest.testTwoServers -- Time elapsed: 3.754 s <<< ERROR!
java.util.ConcurrentModificationException
	at java.base/java.util.ArrayDeque.forEach(ArrayDeque.java:890)
	at io.hyperfoil.impl.CollectingVisitor.visit(CollectingVisitor.java:35)
	at io.hyperfoil.impl.ReflectionAcceptor.accept(ReflectionAcceptor.java:33)
	at io.hyperfoil.impl.CollectingVisitor.visit(CollectingVisitor.java:46)
	at io.hyperfoil.impl.ReflectionAcceptor.accept(ReflectionAcceptor.java:33)
	at io.hyperfoil.impl.CollectingVisitor.visit(CollectingVisitor.java:46)
	at io.hyperfoil.impl.CollectingVisitor.lambda$visit$0(CollectingVisitor.java:35)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1092)
	at io.hyperfoil.impl.CollectingVisitor.visit(CollectingVisitor.java:35)
	at io.hyperfoil.impl.ReflectionAcceptor.accept(ReflectionAcceptor.java:33)
	at io.hyperfoil.impl.CollectingVisitor.visit(CollectingVisitor.java:46)
	at io.hyperfoil.impl.CollectingVisitor.lambda$visit$0(CollectingVisitor.java:35)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1092)
	at io.hyperfoil.impl.CollectingVisitor.visit(CollectingVisitor.java:35)
	at io.hyperfoil.impl.ReflectionAcceptor.accept(ReflectionAcceptor.java:33)
[...]

PR #440, opened 2024-08-30
Fix testTwoServers race condition

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Hyperfoil/issues/439

## Changes proposed

Refactor the test to properly ensure both servers are called and their statistics represent the returned status code

<details>
<summary>
</summary>

</details>


lampajr:
This is actually fixing the problem mainly because the behavior slightly changed as follow:
- Before: server1 /test was waiting on a latch which is later released by server2 /test call (here the server2 call should finish before server1)
- After: the two calls are unrelated, the test simply checks that both are called regardless of the order

Not sure if adding another test trying to replicate the "before" is really what we want or not :thinking:

franz1981 (commented):
Just checked `checkpoint`; good finding!

franz1981 (approved):

Issue #458, opened 2024-09-24
`BaseHttpScenarioTest` tests are randomically failing in the CI


lampajr:
**Describe the bug**

Looks like there are some flaky tests in the CI, see below examples:

https://github.com/Hyperfoil/Hyperfoil/actions/runs/11010920457/job/30573882050?pr=456
```
Error:  Errors: 
Error:    FollowRedirectTest>BaseHttpScenarioTest.lambda$startServer$1:73->BaseHttpScenarioTest.initWithServer:94 » NullPointer Cannot invoke "io.vertx.core.http.HttpServer.actualPort()" because "this.server" is null
Error:    Http1xConnectionStatsTest>BaseHttpScenarioTest.lambda$startServer$1:73->BaseHttpScenarioTest.initWithServer:94 » NullPointer Cannot invoke "io.vertx.core.http.HttpServer.actualPort()" because "this.server" is null
Error:    Http1xConnectionStatsTest>BaseHttpScenarioTest.lambda$startServer$1:73->BaseHttpScenarioTest.initWithServer:94 » NullPointer Cannot invoke "io.vertx.core.http.HttpServer.actualPort()" because "this.server" is null
Error:    Http1xConnectionStatsTest>BaseHttpScenarioTest.lambda$startServer$1:73->BaseHttpScenarioTest.initWithServer:94 » NullPointer Cannot invoke "io.vertx.core.http.HttpServer.actualPort()" because "this.server" is null
```

https://github.com/Hyperfoil/Hyperfoil/actions/runs/11010920457/job/30574072666
```
Error:  Errors: 
Error:    HttpCacheEnablementTest>BaseHttpScenarioTest.lambda$startServer$1:73->BaseHttpScenarioTest.initWithServer:94 » NullPointer Cannot invoke "io.vertx.core.http.HttpServer.actualPort()" because "this.server" is null
```

**To Reproduce**
1. Open a pull request
2. Check the CI workflow

**Expected behavior**
All tests should work, the CI should be green

**Additional context**
The same tests are **working** locally.

lampajr:
Another example:

https://github.com/Hyperfoil/Hyperfoil/actions/runs/11013144176/job/30580831673
```
Error:  Errors: 
[...]

PR #459, opened 2024-09-24
Fix BaseHttpScenarioTest initWithServer

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Hyperfoil/issues/458

## Changes proposed

- [x] Pass the `HttpServer` instance to the `initWithServer` method instead of relying to the global `server` field (which might not be assigned where queried)

<details>
<summary>
</summary>

</details>


lampajr:
@franz1981 this should fix the test failures!

There was a wrong configuration in the base test setup: the `initWithServer` was called in a lambda during the server creation but it was using the server itself to get the port.. which was not good because that global variable might not be initialized yet!

Issue #522, opened 2025-02-19
ThinkTime seems to cause Hyperfoil to hang indefinetly.


iliepandia:
**Describe the bug**

A load works fine until I add `thinkTime` to it. When I add thinkTime to simulate the user looking at the response, the phase will finish, but will never terminate, so I wait indefinitely for it. I can define a maxDuraiton and that will force terminate the phase, but that is not what I want. Also the sessions and connections drop to 0, so I don't understand what is Hyperfoil waiting for.

**To Reproduce**

```
name: micro-load1
http:
  host: http://localhost:8181
  sharedConnections: 1500
phases:
  # Just hit the front end with requests for a simple page
  - userIndexRampUp:
      increasingRate:
        duration: 5s
        maxIterations: 10
        maxSessions: 550
        initialUsersPerSec:
          base: 0
          increment: 5
        targetUsersPerSec:
          base: 5
          increment: 5
        scenario:
          - browse:
              - httpRequest:
                  GET: /payments/
                  sla:
                    - meanResponseTime: 2s
  # Do a buying session ramp up
  # - User logs in
  # - loads the product list
  # - opens a product
  # - places an order for a couple of products
  # - executes a payment
  - userBuyingRampUp:
      increasingRate:
        duration: 6
#        maxDuration: 20 #this was needed because some users hang for some reason!!!
        maxIterations: 10
        maxSessions: 120
        initialUsersPerSec:
          base: 0
          increment: 3
        targetUsersPerSec:
          base: 3
          increment: 3
        scenario:
          - login:
              # load a valid email from a file
              - randomCsvRow:
                  file: emails.txt
                  columns:
                    0: email
              - httpRequest:
                  headers:
                    accept: application/json
                  POST: /users/api/login
                  body:
                    form:
                      - name: email
                        fromVar: email
                      - name: password
                        value: "password"
                  handler:
                    body:
                      # extract the token
                      json:
                        query: ".access_token"
                        toVar: token
                  sla:
                    - meanResponseTime: 2s
          - get-product-list:
              - httpRequest:
                  GET: /products/api/list
                  headers:
                    accept: application/json
                    authorization: Bearer ${token}
                  sla:
                    - meanResponseTime: 2s
              - thinkTime:
                  duration: 1s #user looks at the product for a bit
          - look-at-a-product:
              - httpRequest:
                  GET: /products/api/1
                  headers:
                    accept: application/json
                    authorization: Bearer ${token}
                  sla:
[...]

Issue #377, opened 2024-06-27
Incorrect use of System.currentTimeMillis()


wfouche:
Code like the following appears in Hyperfoil using the System.currentTimeMillis() function to calculate elapsed time.

````
long elapsedMs = System.currentTimeMillis() - absoluteStartTime;
````
But we know  that System.currentTimeMillis() is not a monotonically increasing function, so it may not be used to calculate elasped time. 

https://medium.com/@batuhankadioglu/why-you-should-not-use-currenttimemillis-when-you-calculate-elapsed-time-d65ae36e764b

https://www.javaadvent.com/2023/12/how-not-to-measure-elapsed-time-in-java.html

The correct coding pattern (for measuring elapsed time in milliseconds) is:

````
long beginTimeNanos = System.nanoTime()
.....
long endTimeNanos = System.nanoTime()
long elaspedMS = TimeUnit.NANOSECONDS.toMillis(endTimeNanos - beginTimeNanos)
````

This might just be an oversight because in other areas of Hyperfoil time intervals are correctly calculated, as shown below.

````
long currentTimestamp = System.nanoTime();
long delay = TimeUnit.NANOSECONDS.toMillis(currentTimestamp - lastTimestamp);
````

lampajr:
Hi @wfouche, that's really a good point and the same was raised by @franz1981 as well (which is why with the rate generator refactoring we did [1], we also introduced some additional checks ensuring the elapsed time is not negative - correct me if I am wrong @franz1981).

Furthermore I think this issue is slightly related to https://github.com/Hyperfoil/Hyperfoil/issues/342 as in order to fix that we would have to switch all millis to nanos during the hot path computations. On the other hand, given that https://github.com/Hyperfoil/Hyperfoil/issues/342 is not so straightforward and it might require some additional refactoring _I think we can address this issue separately_ using the approach you suggested (which I agree it seems like the best practice for computing elapsed times).

>When nanoTime is used the variable name only contain the word "time" (and not "timestamp").
Recommending that the same naming convention be considered for Hyperfoil. Therefore, do not use the word timestamp in the name of a variable that contains a nanoTime() time value.

This looks like a good naming convention, I don't see any issue with that so I agree!

[1] https://github.com/Hyperfoil/Hyperfoil/pull/357

franz1981:
[...]

PR #531, opened 2025-03-04
Make delayed step to work with nanos

lampajr:
Related to https://github.com/Hyperfoil/Hyperfoil/issues/522

A bit of context, the only reason for which adding `thinkTime` could make Hyperfoil hanging indefinitely is when https://github.com/Hyperfoil/Hyperfoil/blob/ae0cc9937ca9fa3a25ac737d07c99a5c567e407e/core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java#L65 is properly scheduled after the specified amount of time but for some reason the check at https://github.com/Hyperfoil/Hyperfoil/blob/ae0cc9937ca9fa3a25ac737d07c99a5c567e407e/core/src/main/java/io/hyperfoil/core/steps/AwaitDelayStep.java#L27 returns `false` - this could happen if `System.currentTimeMillis()` returns a value back in time because of some adjustments. Correct me if I am wrong @franz1981. See also this other comment https://github.com/Hyperfoil/Hyperfoil/issues/377#issue-2379044267.

With this PR I am proposing to rely in the `nano` time which should be more reliable

## Changes proposed

- [x] Check delays using nanoTime rather than milliseconds

<details>
<summary>
</summary>

</details>

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/steps/AwaitDelayStep.java
+++ b/core/src/main/java/io/hyperfoil/core/steps/AwaitDelayStep.java
@@ -24,7 +24,8 @@ public class AwaitDelayStep implements Step {
    @Override
    public boolean invoke(Session session) {
       ScheduleDelayStep.Timestamp blockedUntil = (ScheduleDelayStep.Timestamp) key.getObject(session);
-      return System.currentTimeMillis() >= blockedUntil.timestamp;
+      // checking the diff because of the possibility of numerical overflow.
+      return (System.nanoTime() - blockedUntil.timestamp) >= 0;
    }
 
    /**
--- a/core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java
+++ b/core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java
@@ -42,7 +42,7 @@ public class ScheduleDelayStep implements Step, ResourceUtilizer {
    @Override
    public boolean invoke(Session session) {
       Timestamp blockedUntil = (Timestamp) key.activate(session);
-      long now = System.currentTimeMillis();
+      long now = System.nanoTime();
       long baseTimestamp;
       switch (type) {
          case FROM_LAST:
@@ -62,7 +62,7 @@ public class ScheduleDelayStep implements Step, ResourceUtilizer {
       long delay = blockedUntil.timestamp - now;
       if (delay > 0) {
          log.trace("Scheduling #{} to run in {}", session.uniqueId(), delay);
-         session.executor().schedule(session.runTask(), delay, TimeUnit.MILLISECONDS);
+         session.executor().schedule(session.runTask(), delay, TimeUnit.NANOSECONDS);
       } else {
          log.trace("Continuing, duration {} resulted in delay {}", duration, delay);
       }
@@ -126,7 +126,7 @@ public class ScheduleDelayStep implements Step, ResourceUtilizer {
       }
 
       public Builder duration(long duration, TimeUnit timeUnit) {
-         this.duration = timeUnit == null ? 0 : timeUnit.toMillis(duration);
+         this.duration = timeUnit == null ? 0 : timeUnit.toNanos(duration);
          return this;
       }
 
@@ -137,7 +137,7 @@ public class ScheduleDelayStep implements Step, ResourceUtilizer {
        * @return Self.
        */
       public Builder duration(String duration) {
-         this.duration = Util.parseToMillis(duration);
+         this.duration = Util.parseToNanos(duration);
          return this;
       }
 
--- a/test-suite/src/test/java/io/hyperfoil/test/PacingTest.java
+++ b/test-suite/src/test/java/io/hyperfoil/test/PacingTest.java
@@ -6,7 +6,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 
@@ -16,7 +15,6 @@ import io.hyperfoil.http.api.HttpMethod;
 import io.vertx.junit5.VertxExtension;
 
 @ExtendWith(VertxExtension.class)
-@Disabled // Ignoring in default run since it takes 3x 5 seconds
 public class PacingTest extends BaseHttpScenarioTest {
    @Override
    protected void initRouter() {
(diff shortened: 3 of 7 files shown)


lampajr:
I still think that there are no tests ensuring the specified delay is actually passed, need to check whether we can easily add a test like this

test-suite/src/test/java/io/hyperfoil/test/PacingTest.java
```diff
@@ -45,11 +43,19 @@ public void testThinkTimes() {
...
-      assertThat(snapshot.requestCount).isEqualTo(expected);
-      assertThat(snapshot.responseCount).isEqualTo(expected);
+   @Test
+   public void testSmallThinkTimes() {
+      scenario().initialSequence("loop")
+            .step(SC).loop("counter", 5)
+            .steps()
+            .step(SC).httpRequest(HttpMethod.GET).path("/test").endStep()
+            .step(SC).clearHttpCache()
+            .step(SC).thinkTime(500, TimeUnit.NANOSECONDS).endStep()
```
franz1981, line 53: you can use 1 as well if you think is valuable ;)

core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java
```diff
@@ -298,44 +298,23 @@ public ThinkTimeBuilder init(String param) {
...
-      private final long duration;
-      private final long max;
-      private final long min;
-
-      RandomNegExpDuration(long duration, long max, long min) {
-         this.duration = duration;
-         this.max = max;
-         this.min = min;
-      }
+   private record RandomNegExpDuration(long duration, long max, long min) implements SerializableToLongFunction<Session> {
```
franz1981, line 317: We do perform reflective accesses or any other serialization we expect to break making this a `record`?

core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java
```diff
@@ -298,44 +298,23 @@ public ThinkTimeBuilder init(String param) {
...
 
-   private static class RandomLinearDuration implements SerializableToLongFunction<Session> {
-      private final long min;
-      private final long max;
-
-      RandomLinearDuration(long min, long max) {
-         this.min = min;
-         this.max = max;
-      }
+   private record RandomLinearDuration(long min, long max) implements SerializableToLongFunction<Session> {
```
franz1981, line 309: We do perform reflective accesses or any other serialization we expect to break making this a record?

core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java
```diff
@@ -126,7 +126,7 @@ public Builder key(String key) {
       }
 
       public Builder duration(long duration, TimeUnit timeUnit) {
-         this.duration = timeUnit == null ? 0 : timeUnit.toMillis(duration);
+         this.duration = timeUnit == null ? 0 : timeUnit.toNanos(duration);
```
franz1981, line 129: we should check that is not negative - but TBH is a non-problem usually

core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java
```diff
@@ -62,7 +62,7 @@ public boolean invoke(Session session) {
       long delay = blockedUntil.timestamp - now;
       if (delay > 0) {
          log.trace("Scheduling #{} to run in {}", session.uniqueId(), delay);
-         session.executor().schedule(session.runTask(), delay, TimeUnit.MILLISECONDS);
+         session.executor().schedule(session.runTask(), delay, TimeUnit.NANOSECONDS);
```
franz1981, line 65: This is not saving the ScheduledFuture nowhere - it is just fixing the ns granularity right?

core/src/main/java/io/hyperfoil/core/steps/AwaitDelayStep.java
```diff
@@ -24,7 +24,7 @@ public AwaitDelayStep(ReadAccess key) {
    @Override
    public boolean invoke(Session session) {
       ScheduleDelayStep.Timestamp blockedUntil = (ScheduleDelayStep.Timestamp) key.getObject(session);
-      return System.currentTimeMillis() >= blockedUntil.timestamp;
+      return System.nanoTime() >= blockedUntil.timestamp;
```
franz1981, line 27: compute the distance here - since System.nanoTime can wrap (negative to positive) - the diff is always safe due to math oveflow - see doc at https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--

core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java
lampajr: Yeah correct, this is just about the `ns` granularity

core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java
lampajr: This check is already performed as part of the step's `build()`, e.g., https://github.com/Hyperfoil/Hyperfoil/blob/ae0cc9937ca9fa3a25ac737d07c99a5c567e407e/core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java#L228

core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java
lampajr: revert just to be sure as not really needed here

core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java
lampajr: revert just to be sure as not really needed here

core/src/main/java/io/hyperfoil/core/steps/AwaitDelayStep.java
lampajr: Thanks for the suggestion! Updated and added a comment as well

core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java
lampajr: yeah I see that, but I am expecting that when you call the `build()` on the `ScheduledDelayStep` it will be found that is negative here https://github.com/Hyperfoil/Hyperfoil/blob/ae0cc9937ca9fa3a25ac737d07c99a5c567e407e/core/src/main/java/io/hyperfoil/core/steps/ScheduleDelayStep.java#L224-L231

This is going to be called _after_ the duration is set (and converted)

franz1981 (approved):
LGTM! Well done!

Issue #536, opened 2025-03-18
Calendar::getInstance allocates a new Calendar instance while parsing dates


franz1981:
Http caching requires to parse few dates from the headers,  creating a costly `Calendar` instance in the hot path, see

![Image](https://github.com/user-attachments/assets/4210a3b6-2069-4150-9478-e91fa704e9d4)

As commented at https://github.com/Hyperfoil/Hyperfoil/blob/c637f10cc4f73f9f3b6eb6f06bbfd52b51ff44d9/http/src/main/java/io/hyperfoil/http/HttpUtil.java#L144-L147

this should be better avoided, either by saving in an map/array per calendar type bounded to the current thread (i.e. using `static final FastThreadLocal<Calendars>` in `HttpUtil`).

Since Hyperfoil containers a JMH module it would be possible to verify the performance improvement there.

PR #538, opened 2025-03-18
Reduce GC pressure while parsing HTTP headers (Fixes #537)

### LINKED ISSUE #537: Reduce GC pressure while parsing HTTP headers
Both HTTP 1.1 names and values allocates in the hot path, see

Image

This could be improved by uring a map/trie to cache the most common expected ones or just save any allocations if there are no HTTP header handlers registered.

franz1981:
It saves handling the headers (parsing and materialize them as proper `AsciiString`/`String`s pairs) - if not required.
This will apply if HTTP caching is disabled (e.g. wrk/wrk2) since HTTP caching requires parsing dates from proper headers (sees #536) - which means that only expert/informed users would benefit of it.

Fixes #537 - although there's still room to reduce the pressure with proper HTTP header names/value caching.

### DIFF
--- a/http/src/main/java/io/hyperfoil/http/api/HttpResponseHandlers.java
+++ b/http/src/main/java/io/hyperfoil/http/api/HttpResponseHandlers.java
@@ -6,6 +6,8 @@ import io.netty.buffer.ByteBuf;
 public interface HttpResponseHandlers extends ResponseHandlers<HttpRequest> {
    void handleStatus(HttpRequest request, int status, String reason);
 
+   boolean requiresHandlingHeaders(HttpRequest request);
+
    void handleHeader(HttpRequest request, CharSequence header, CharSequence value);
 
    void handleBodyPart(HttpRequest request, ByteBuf data, int offset, int length, boolean isLastPart);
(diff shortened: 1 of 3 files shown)


lampajr (approved):

franz1981:
As a side note: 
https://github.com/Hyperfoil/Hyperfoil/blob/965dedefbdd3472a8abdaf93077b00406c0d27f8/http/src/main/java/io/hyperfoil/http/HttpCacheImpl.java#L413-L464

shows under which cases we make uses of HTTP caching, but sadly `Date` is fairly common - and requires the other headers (Cache-Control, Expires, Age) to not be specified.
We can think about having a fast-path which check for these header(s) presence and perform a fast check (with parsing) sole of the `Date` one - straight from the `ByteBuf` - this would save using the HTTP cache (and parsing fully the headers - creating any object) , if not needed, but looks very error prone.
I still prefer users to consciously decide to give up on HTTP caching via our ergonomics, but is still good to know that there are other options too, to improve it.

franz1981:
@lampajr in theory we still check the header name/value boundaries and I have the feeling that we could skip them too, in case http caching is disabled - I have to double check

lampajr:
> @lampajr in theory we still check the header name/value boundaries and I have the feeling that we could skip them too, in case http caching is disabled - I have to double check

I am not sure about that, this is a flamegraph when http cache is disabled:
image

And we are just parsing the headers from `readHeaders` but I think this is required as users can add their own header handlers even if the HTTP cache is disabled

PR #539, opened 2025-03-19
Allocation free date parsing

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Hyperfoil/issues/536

## Changes proposed

Create a static map in `FastThreadLocal` with the most common TZ Calendars (i.e., `UTC` and `GMT`) so that we don't need to allocate such Calendars object in the hot path (i.e., when parsing the dates)

<details>
<summary>
</summary>

</details>

### DIFF
--- a/http/src/main/java/io/hyperfoil/http/HttpUtil.java
+++ b/http/src/main/java/io/hyperfoil/http/HttpUtil.java
@@ -3,6 +3,7 @@ package io.hyperfoil.http;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Calendar;
+import java.util.Map;
 import java.util.TimeZone;
 
 import org.apache.logging.log4j.LogManager;
@@ -11,13 +12,27 @@ import org.apache.logging.log4j.Logger;
 import io.hyperfoil.impl.Util;
 import io.netty.buffer.ByteBuf;
 import io.netty.util.AsciiString;
+import io.netty.util.concurrent.FastThreadLocal;
 
 public final class HttpUtil {
    private static final Logger log = LogManager.getLogger(HttpUtil.class);
+
+   private static final AsciiString UTC_ASCII = new AsciiString("UTC");
+   private static final AsciiString GMT_ASCII = new AsciiString("GMT");
+   private static final TimeZone UTC = TimeZone.getTimeZone(UTC_ASCII.toString());
+   private static final TimeZone GMT = TimeZone.getTimeZone(GMT_ASCII.toString());
+   private static final FastThreadLocal<Map<AsciiString, Calendar>> CALENDARS = new FastThreadLocal<>() {
+      @Override
+      protected Map<AsciiString, Calendar> initialValue() {
+         return Map.of(
+               UTC_ASCII, Calendar.getInstance(UTC),
+               GMT_ASCII, Calendar.getInstance(GMT));
+      }
+   };
+
    private static final CharSequence[] MONTHS = { "jan", "feb", "mar", "apr", "may", "jun", "jul", "aug", "sep", "oct", "nov",
          "dec" };
-   private static final TimeZone UTC = TimeZone.getTimeZone("UTC");
-   private static final TimeZone GMT = TimeZone.getTimeZone("GMT");
+
    private static final byte[] BYTES_80 = "80".getBytes(StandardCharsets.UTF_8);
    private static final byte[] BYTES_443 = "443".getBytes(StandardCharsets.UTF_8);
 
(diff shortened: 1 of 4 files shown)


http/src/main/java/io/hyperfoil/http/HttpUtil.java
```diff
@@ -11,13 +12,25 @@
 import io.hyperfoil.impl.Util;
 import io.netty.buffer.ByteBuf;
 import io.netty.util.AsciiString;
+import io.netty.util.concurrent.FastThreadLocal;
 
 public final class HttpUtil {
+   private static final String UTC_STR = "UTC";
```
franz1981, line 18: Use AsciiString here if the seq tends to be an AsciiString

lampajr: Done, even if I adon't think there is much impact

lampajr:
@franz1981 not sure whether this is how you were thinking the fix would have been but I think it is actually improving that parsing

http/src/main/java/io/hyperfoil/http/HttpUtil.java
```diff
@@ -131,18 +146,21 @@ public static long parseDate(CharSequence seq, int begin, int end) {
...
          ; // skip spaces
 
-      TimeZone timeZone = UTC;
+      Calendar calendar;
       if (i < end) {
-         if (end - i >= 3 && AsciiString.regionMatches(seq, false, i, "GMT", 0, 3)) {
-            timeZone = GMT;
-         } else if (end - i >= 3 && AsciiString.regionMatches(seq, false, i, "UTC", 0, 3)) {
-            timeZone = UTC;
+         if (end - i >= 3 && AsciiString.regionMatches(seq, false, i, GMT_ASCII, 0, 3)) {
```
franz1981, line 151: you can use `AsciiString::equals` here if the caller is using `0` and `seq.length()`

franz1981 (changes requested):

http/src/main/java/io/hyperfoil/http/HttpUtil.java
lampajr: > you can use AsciiString::equals here if the caller is using 0 and seq.length() , since it is not case-insensitive.

I am not sure I got this. I can use equals if and only if the `end - i == 3` but this does not cover the case when `end - i > 3` as in this case I would need to compare GMT/UTC AsciiString with the  subset 0-3 of the input `seq`.

lampajr: and in all cases I need to extract the subsequence from `i`

lampajr: It would then become something like this, not sure that's how you were thinking
```java
         if (end - i == 3) {
            if (GMT_ASCII.equals(seq.subSequence(i, i+3))) {
               calendar = CALENDARS.get().get(GMT_ASCII);
            } else if (GMT_ASCII.equals(seq.subSequence(i, i+3))) {
               calendar = CALENDARS.get().get(UTC_ASCII);
            }
         } else {
            if (AsciiString.regionMatches(seq, false, i, GMT_ASCII, 0, 3)) {
               calendar = CALENDARS.get().get(GMT_ASCII);
            } else if (AsciiString.regionMatches(seq, false, i, UTC_ASCII, 0, 3)) {
               calendar = CALENDARS.get().get(UTC_ASCII);
            }
         }
         
         if (calendar == null) {
            // allocate a new calendar only if not UTC or GMT
            TimeZone timeZone = TimeZone.getTimeZone(seq.subSequence(i, end).toString());
            calendar = Calendar.getInstance(timeZone);
         }
```

franz1981: > I can use equals if and only if the end - i == 3 but this does not cover the case when end - i > 3

stupid me, I have misread it

franz1981 (approved):
LGTM well done 👍

http/src/main/java/io/hyperfoil/http/HttpUtil.java
franz1981: Only if are the most used there 🙏 you can verify by debugging

franz1981:
Re your changes to use `regionMatches` I've noticed Netty isn't doing the "right thing" and sent https://github.com/netty/netty/pull/14947 to fix it :pray:

lampajr:
> Re your changes to use `AsciiString::regionMatches` I've noticed Netty isn't doing the "right thing" and sent [netty/netty#14947](https://github.com/netty/netty/pull/14947) to fix it 🙏

Yeah I noticed that it was not implementing a fast path if comparing AsciiString with AsciiString

franz1981:
yeah but I still don't get what we allocate there - so try running it with
```
-prof "async:event=alloc;output=html;dir=/tmp;libPath=${ASYNC_PROFILER_HOME}/lib/libasyncProfiler.so" 
```
using async-profiler 3.0 and setting properly the env var there.
So we can see what's been allocated there - which is still way too much to me

lampajr:
> yeah but I still don't get what we allocate there - so try running it with
> 
> ```
> -prof "async:event=alloc;output=html;dir=/tmp;libPath=${ASYNC_PROFILER_HOME}/lib/libasyncProfiler.so" 
> ```
> 
> using async-profiler 3.0 and setting properly the env var there. So we can see what's been allocated there - which is still way too much to me

Keep in mind that one of the Date string that we are passing to the test is **NOT** in the cached map (i.e., `CET`) as we are only caching `UTC` and `GMT` - so I think that some allocation there is reasonable:

image

PR #540, opened 2025-03-19
Optimized Content-Length and Transfer-Encoding matches

franz1981:
It performs earlier the logic of detecting header name/value limits, to speed up (and eventually, save) matching `Content-Length` and `Transfer-Encoding` headers vs the raw `ByteBuf`.
This enable skipping the whitespaces while performing the match itself, since it is performed earlier, regardless.

## Check List


franz1981:
This has been discovered while fixing https://github.com/Hyperfoil/Hyperfoil/issues/537

lampajr (approved):
LGTM

PR #541, opened 2025-03-21
Stop using a Session if the delay is already expired

franz1981:
## Changes proposed

Currently, due to lack of cpu resources (e.g. on containers, GC, etc etc) a short delay can cause a scheduled task to use a Session which it doesn't "own" anymore, because released after executing the sequence it belongs.
This is happening because:
- `ScheduleDelayStep::invoke` schedule a short living scheduled task
- `AwaitDelayStep` is always executed after `ScheduleDelayStep`, and due to many reasons, can find the delay to be already expired
- `Session`s completes and got used elsewhere
- the short living scheduled task executes `Session::runTask`  on a `Session` which doesn't "own" anymore

We're not aware of any adverse effect of using a `Session` which we don't own - since `Session`s are single-threaded, at worse it can run some stucked `step` once more, while making it to progress earlier than expected, which means that it depends by the nature of the step to be executed.

## Check List

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/steps/AwaitDelayStep.java
+++ b/core/src/main/java/io/hyperfoil/core/steps/AwaitDelayStep.java
@@ -25,7 +25,18 @@ public class AwaitDelayStep implements Step {
    public boolean invoke(Session session) {
       ScheduleDelayStep.Timestamp blockedUntil = (ScheduleDelayStep.Timestamp) key.getObject(session);
       // checking the diff because of the possibility of numerical overflow.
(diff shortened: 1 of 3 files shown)


franz1981:
This work is triggered by some investigations around https://github.com/Hyperfoil/Hyperfoil/issues/522 and a follow up to https://github.com/Hyperfoil/Hyperfoil/pull/531 to have a "cleaned up" `ThinkTime` feature - as it was missing this last bit.

IDK if @iliepandia wish to give it a shot without using the proposed `awaitVar` which fixed it last time :pray:

lampajr (approved):
LGTM

PR #556, opened 2025-05-19
Add SSL handshake timeout setting

yawkat:


lampajr (approved):
LGTM!

lampajr:
> LGTM!

There is still one issue, I was trying it but it looks like the configuration is not actually considered. Need to look deeper 👀

yawkat:
Hmm I tried with the programmatic configuration and it worked fine. I did not try the yaml config

lampajr:
> Hmm I tried with the programmatic configuration and it worked fine. I did not try the yaml config

yeah me too, I am exactly looking at the yaml config now

lampajr (approved):
Ok it works!

PR #559, opened 2025-05-22
Disable HTTP cache by default to save memory

lampajr:
## Changes proposed

I would propose to disable the HTTP cache by default in all benchmarks.

From my understanding most of use cases do not use the HTTP cache and given that creating HTTP cache is quite resource consuming, I am proposing to disable it by default.

Users can still enable it by adding `useHttpCache: true` in the `http` entry.

<details>
<summary>
</summary>

</details>

### DIFF
--- a/docs/site/content/en/docs/user-guide/benchmark/http.md
+++ b/docs/site/content/en/docs/user-guide/benchmark/http.md
@@ -28,25 +28,25 @@ http:
 
 HTTP configuration has these properties:
 
-| Property          | Default | Description |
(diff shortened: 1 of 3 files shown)


lampajr:
@franz1981 wdyt about this?

franz1981:
IDK, I think @rvansa implemented it cause Gatling have it and it's enabled by default.
If people move our from Gatling to Hyperfoil, I wish them to be aware that's a key diff
If we are happy with it, I am happy as well since other load gen doesn't do it afaik (wrk2, can you check?)

PR #567, opened 2025-05-27
Add option to limit max length in downloadLog

yawkat:
When the agent has a lot of errors, downloading the logs can take a long time, leading to timeouts. This PR adds an option to the download endpoint to set a maximum size.

I tested this with SSH agents but I don't have a k8s test setup.

### DIFF
--- a/k8s-deployer/src/main/java/io/hyperfoil/deploy/k8s/K8sDeployer.java
+++ b/k8s-deployer/src/main/java/io/hyperfoil/deploy/k8s/K8sDeployer.java
@@ -306,25 +306,26 @@ public class K8sDeployer implements Deployer {
    }
 
    @Override
-   public void downloadControllerLog(long offset, String destinationFile, Handler<AsyncResult<Void>> handler) {
-      downloadRunningLog(CONTROLLER_POD_NAME, offset, destinationFile, handler);
+   public void downloadControllerLog(long offset, long maxLength, String destinationFile, Handler<AsyncResult<Void>> handler) {
+      downloadRunningLog(CONTROLLER_POD_NAME, offset, maxLength, destinationFile, handler);
    }
 
    @Override
-   public void downloadAgentLog(DeployedAgent deployedAgent, long offset, String destinationFile,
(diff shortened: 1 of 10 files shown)


franz1981 (changes requested):

k8s-deployer/src/main/java/io/hyperfoil/deploy/k8s/K8sDeployer.java
```diff
@@ -306,25 +306,26 @@ public boolean hasControllerLog() {
...
-   public void downloadAgentLog(DeployedAgent deployedAgent, long offset, String destinationFile,
+   public void downloadAgentLog(DeployedAgent deployedAgent, long offset, long maxLength, String destinationFile,
          Handler<AsyncResult<Void>> handler) {
       K8sAgent agent = (K8sAgent) deployedAgent;
       ensureClient();
       if (agent.outputPath != null) {
          try (InputStream stream = new FileInputStream(agent.outputPath.toFile())) {
             skipBytes(offset, stream);
-            Files.copy(stream, Paths.get(destinationFile), StandardCopyOption.REPLACE_EXISTING);
+            Files.copy(new TruncatingInputStream(stream, maxLength), Paths.get(destinationFile),
```
franz1981, line 321: Similarly to before: can you add a log in case the copied data is less than the existing file size?

yawkat:
@franz1981 I think it's better for the client to handle this case. I can imagine non-error uses for this API too, for example some form of rudimentary streaming

franz1981:
@yawkat wdym?

franz1981:
ping @yawkat 

> I think it's better for the client to handle this case

I don't understand if these changes are good or not for you :pray:

yawkat:
Basically, I think that taking a limited chunk of a large log is not necessarily an error that deserves a log message. It could be used to stream a normal log in small chunks too. It's up to the client whether exceeding the size limit is an error or not.

franz1981:
Got it an agree thanks for the pr @yawkat

Issue #362, opened 2024-06-11
Upgrade to JDK 21


lampajr:
Upgrade Hyperfoil to JDK 21

franz1981:
Check https://bugs.openjdk.org/browse/JDK-8336492 as well: it could be related (or not) or just an additional reason why we shouldn't rely on lambda serializated form (as stable)

PR #573, opened 2025-07-11
Upgrade to JDK 21

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Hyperfoil/issues/362

## Changes proposed

- [x] Upgrade from JDK 17 to 21

<details>
<summary>
</summary>

</details>

### DIFF
--- a/api/src/main/java/io/hyperfoil/api/config/BuilderBase.java
+++ b/api/src/main/java/io/hyperfoil/api/config/BuilderBase.java
@@ -86,7 +86,7 @@ public interface BuilderBase<S extends BuilderBase<S>> {
    default S copy(Object newParent) {
       if (getClass().isSynthetic()) {
          // This is most likely a lambda supplier of the instance (which should be immutable anyway)
-         assert getClass().getSimpleName().contains("$$Lambda$");
+         assert getClass().getSimpleName().contains("$$Lambda");
          return (S) this;
       }
       try {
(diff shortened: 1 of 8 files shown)


api/src/main/java/io/hyperfoil/api/config/BuilderBase.java
```diff
@@ -86,7 +86,7 @@ private void tryPrepare(Class<?> clz, String name, Class<?> type, Object value)
    default S copy(Object newParent) {
       if (getClass().isSynthetic()) {
          // This is most likely a lambda supplier of the instance (which should be immutable anyway)
-         assert getClass().getSimpleName().contains("$$Lambda$");
+         assert getClass().getSimpleName().contains("$$Lambda");
```
lampajr, line 89: Looks like something changed on how the simple name is composed in case of lambdas, still need to confirm this

lampajr: I can confirm this changed happened in the JDK 21:
* Related JDK issue: https://bugs.openjdk.org/browse/JDK-8292914
* Commit: https://github.com/openjdk/jdk/commit/b527edd3388ad6a0d44a291983b08b2b5c023f8f

```
        }
-        return name.replace('.', '/') + "$$Lambda$" + counter.incrementAndGet();
+        return name.replace('.', '/') + "$$Lambda";
    }
```

franz1981: So we don't need to have a Runtime version check to pick the right way to name them?
Or apply some static check on a know lambda name and decide the strategy?
From my understanding relying on how lambdas are named can be...brittle

franz1981: Regardless, well done to have found the issue eh : ❤️

Issue #589, opened 2025-08-08
Add max/p99.99/p99.999 to reports


gunnarmorling:
As per discussion https://github.com/Hyperfoil/Hyperfoil/discussions/587

gunnarmorling:
Thanks for the quick experiment, @franz1981! Personally, I think having p99.99 would be nice; but agreed that having the HDR Histogram download (#588) will cover all cases (btw. its web viewer lets you dynamically choose btw. the maximum latency you want to display), so I'm happy to have this one closed in favor of #588.

diegolovison:
https://github.com/Hyperfoil/report/pull/18 add the extreme percentiles (p99.99, p99.999, max) requested in via a new Key Performance Percentiles table. This provides the exact data @gunnarmorling requested without introducing the visual noise or Y-axis scaling issues on the time-series charts that @franz1981 noted.

PR #590, opened 2025-08-08
Changing percentiles to save computing their percentage value

franz1981:
While working on #589 I've found rather error-prone to add a single percentile class, causing several math errors due to rounding the percentile * 100,
This change aim to fix this while adding few tests to verify there is no impact in the changes.


lampajr (approved):
LGTM

PR #640, opened 2025-12-22
Shutdown should wait until finished

diegolovison:
Shutdown should wait until finished


franz1981 (commented):
Check  my comment on the caller thread

PR #653, opened 2026-01-15
Assert for the wrk command result

diegolovison:
Assert for the wrk command result


franz1981 (approved):
Just to make sure it would work, can you introduce a bug in the server and see if it makes the test to fail ?

diegolovison:
Merging because it was approved.

PR #654, opened 2026-01-15
Add debug messages for helping troubleshooting cluster issues

diegolovison:
Add debug messages for helping troubleshooting cluster issues


franz1981 (commented):
Make sure to prepend with `isDebugEnabled` or similar if the code path is an hot/frequently hit one (which shouldn't)

PR #663, opened 2026-01-22
Add warmupDuration option

diegolovison:
Per my understanding, there is no calibration phase on https://github.com/wg/wrk/blob/master/src/wrk.c

### DIFF
--- a/test-suite/src/test/java/io/hyperfoil/scenario/WrkScenarioTest.java
+++ b/test-suite/src/test/java/io/hyperfoil/scenario/WrkScenarioTest.java
@@ -50,7 +50,7 @@ public class WrkScenarioTest extends BaseWrkBenchmarkTest {
    @Test
    public void testWrk2() throws URISyntaxException {
 
-      String url = "localhost:" + httpServer.actualPort() + "/foo/bar";
+      String url = "localhost:" + httpServer.actualPort() + "/highway";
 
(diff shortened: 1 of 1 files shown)


diegolovison:
> Check why that phase was there before removing it

Based on git history: https://github.com/Hyperfoil/Hyperfoil/commit/fef405ad04bf4683e7e5860823b2facfd332ab84#diff-629410907e4ac7a18a4c187094272449bfb3bf1f20a8835e688630e1e7ceca5cR140-R141
It seems like it was added by mistake because there is no reference for the calibration phase on https://github.com/wg/wrk/blob/master/src/wrk.c

> Not only, wrk.c is a native app...

May you please explain?

diegolovison:
Our goal with this command is to be a drop-in replacement for the standard wrk CLI, including a calibration phase, which actually makes us **incompatible**. Users running a 5s test (like -d 5s) expect a 5s execution; if we force a calibration phase, the test takes significantly longer or behaves unpredictably compared to the C version.

I do understand your point, but the way to fix it is exactly what @johnaohara suggested and **not introducing a calibration phase**.
- build a native client binary
- share an aot library with pre-compiled code

franz1981:
Thanks @lampajr and hi!

@diegolovison there is already an opened issue re AOT by me iirc but we cannot, in the meantime, and in absence of a proper replacement, drop it.
But..we can add an option to skip it so people which have a small load can trust that the number of requests is what they expect to be (and I remember I have opened an issue for this too)

diegolovison:
> I vote for the opposite.

Ok. Going to add `--skip-calibration` option

PR #665, opened 2026-01-23
Refactor phase lifecycle and statistics collection for improved reliability

diegolovison:
Refactor phase lifecycle and statistics collection for improved reliability

- Extract connection opening logic into waitOpenConnection() method for better separation of concerns
- Ensure statistics executor properly waits for all pending tasks before shutdown
- Add sequential task execution guarantee with single-threaded executor documentation
- Refactor phaseChanged to use onPhaseChanged hook pattern for extensibility
- Fix phase name reference from phase.name to phase.name() for consistency
- Ensure statistics are fully collected before termination with configurable timeout

This PR basically moves things around and adds asserts. There is a misbehavior in `LocalSimulationRunner` and it is not matching what `AgentVerticle` is doing. I know the issue exists, but I still don't have a concrete reproducer for it.

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/impl/LocalSimulationRunner.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/LocalSimulationRunner.java
@@ -1,7 +1,6 @@
 package io.hyperfoil.core.impl;
 
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


franz1981 (approved):
Thanks ❤️

PR #690, opened 2026-02-16
CounterProcessor for troubleshooting

diegolovison:
CounterProcessor for troubleshooting

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,30 @@ Follow these steps:
    ```
    You should be able to connect without entering a password.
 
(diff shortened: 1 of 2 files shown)


franz1981:
This doesn't seem needed as you could do something similar with the existing components e.g. counters and logging

diegolovison:
Run `jbang wrk2@hyperfoil -t1 -c2 -d10s -R500 --latency http://localhost:8080` and look at `/tmp/hyperfoil/run/0000/all.json`

PR #692, opened 2026-02-17
Allows load generator to use nano second granularity

### LINKED ISSUE #342: OpenModelPhase cannot go past 1000 rps

franz1981:
## Fixes Issue

Fixes #342 

I have to check this against https://github.com/franz1981/Netty-VirtualThread-Scheduler/pull/81 
i.e. if we uses native transport having nano seconds granularity will works well, but what happen if it doesn't?

### DIFF
--- a/core/src/test/java/io/hyperfoil/core/impl/rate/PoissonRampRateGeneratorTest.java
+++ b/core/src/test/java/io/hyperfoil/core/impl/rate/PoissonRampRateGeneratorTest.java
@@ -23,13 +23,16 @@ public class PoissonRampRateGeneratorTest extends RateGeneratorTest {
 
    @Override
    void assertSamplesWithoutSkew(final double[] samples, final long totalUsers) {
-      final double[] interArrivalTimes = computeInterArrivalTimes(samples);
-      final double[] fireTimesOnIntervals = new double[interArrivalTimes.length];
-      double elapsedTime = 0;
-      for (int i = 0; i < interArrivalTimes.length; i++) {
-         final double rpNs = computeRateAtTime(1.0 / 1_000_000_000.0, 10.0 / 1_000_000_000.0, 10_000_000_000L, elapsedTime);
-         fireTimesOnIntervals[i] = interArrivalTimes[i] * rpNs;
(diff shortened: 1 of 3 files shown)


core/src/test/java/io/hyperfoil/core/impl/rate/NanosecondBurstinessTest.java
```diff
@@ -0,0 +1,259 @@
...
+      for (int i = 1; i < samples; i++) {
+         long currentMs = fireTimes[i] / 1_000_000;
+         if (currentMs != prevMs) {
+            distinctMillisValues++;
+            prevMs = currentMs;
+         }
+      }
+      // 100 samples at 100,000 ns apart = 10,000,000 ns = 10 ms total span
+      // So we should see roughly 10 distinct ms values
+      assertTrue(distinctMillisValues >= 9,
```
diegolovison: We are testing only the lower bound. How about the upper bound?

diegolovison:
This PR LGTM, but if you agree, I would like to have a new commit before the merge.

Today, the values are represented as `9_000_000_000L`, `100_000L`, and so on. Without looking within the code, you don't understand what it represents.
My suggestion is to change to `Duration` within the tests and make the constructors receive `Duration` as well.
Internally, it should be converted to `long`, then the hotpath will work only with long.

```java
// Instead of:
RateGenerator.rampRate(1, 10, 10_000_000_000L)

// Use:
RateGenerator.rampRate(1, 10, Duration.ofSeconds(10))
```

Another note, do we need the following scenarios?
- Very low (<1000)
- very high (>100k)

core/src/test/java/io/hyperfoil/core/impl/rate/NanosecondBurstinessTest.java
franz1981: We have other tests which looks to the reported values singularly which should behave as expected (accordly to the function described) and we test the distance too

diegolovison (approved):

PR #696, opened 2026-02-19
Fix internal coordinated omission for open model phases

franz1981:
Fixes #592 

It exposes a new ergonomic (used by default on `Wrk2` wrapper - which can be very high rate - and use the computed `intended start time ns` for the response times computation, adjusting the one collected during the preparation http step, which could be prone to internal delays or GC/throttling.

This erognomics could be enabled on any other test with the limit that the http request step should be the nearest one to the rate tick generation (if I got it right!).

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
@@ -104,7 +104,7 @@ public abstract class PhaseInstanceImpl implements PhaseInstance {
       recordAbsoluteStartTime();
       log.debug("{} changing status to RUNNING", def.name);
       phaseChangeHandler.onChange(def, Status.RUNNING, false, error)
-            .thenRunAsync(this::proceed, executorGroup.next());
+            .thenRun(this::proceed);
    }
 
    protected abstract void proceed();


core/src/main/java/io/hyperfoil/core/session/SessionImpl.java
```diff
@@ -53,6 +53,8 @@ class SessionImpl implements Session {
    private AgentData agentData;
    private GlobalData globalData;
    private SessionStatistics statistics;
+   private long startTimestampMillis;
```
franz1981: probably i should have used exactly the same name prefix `intended`

franz1981:
Thinking about it twice, only throttled users would be unable to correctly uses the intended start time: I have to think if I could address this as it is or in a different way

franz1981:
I have to check the duration of some new tests since they look too long IMHO

eg
```
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 9.236 s -- in io.hyperfoil.http.CompensateInternalLatencyTest
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4.306 s -- in io.hyperfoil.http.ThrottledSessionsTest

```

Issue #694, opened 2026-02-18
CPu watchdog assume a 10ms tick while modern kernels have 100ms


franz1981:
This lead to make the cpu watchdog to complain for no reasons...it would be great if we could get that value out fo the system to make the cpu watchdog to work more reliably

PR #697, opened 2026-02-19
Resolve CpuWatchdog false positives due to tick rate assumptions

diegolovison:
Fixes: https://github.com/Hyperfoil/Hyperfoil/issues/694

The original CpuWatchdog was experiencing false positives because it calculated CPU usage by comparing kernel ticks against wall-clock time (System.nanoTime()). It assumed a hardcoded 10ms tick (TICK_NANOS = 10_000_000).

If the kernel skips ticks to save power (tickless kernels / NOHZ), or if our Java thread experiences minor scheduling jitter, comparing elapsed nanoseconds to kernel ticks causes severe math drift, falsely triggering the 100% CPU usage alarm.


franz1981 (approved):
I really like this approach, which is similar to what `top` does AFAIK, and much better: keep the ratios and no need for jiffies ;)

Issue #678, opened 2026-01-29
Verify native transport usage for nanosecond timing precision (pwait2 syscall)


diegolovison:
## Context

As discussed in https://github.com/franz1981/Netty-VirtualThread-Scheduler/pull/81, using Netty's native transports (Epoll on Linux, KQueue on macOS) is critical for accurate timing measurements in benchmarking tools. Without native transports, timing resolution may be limited to millisecond granularity instead of nanoseconds.

The key indicator is the presence of `pwait2` syscalls, which provide nanosecond-level precision for timing operations.

Hyperfoil already has native transport support implemented in `EventLoopFactory`. It needs to be double-checked.

diegolovison:
@franz1981 I would like to double-check if I understood you correctly:
- retrieve if we have epoll_pwait2 on the host machine
- if not, create a WARN message

Is my understanding accurate?

PR #700, opened 2026-02-20
Add warn message about sub-millisecond and timing precision

diegolovison:
Fixes: https://github.com/Hyperfoil/Hyperfoil/issues/678

Add warn message about sub-millisecond and timing precision


franz1981 (approved):
LGTM!
What it says on your machine?

Issue #533, opened 2025-03-14
JsonParser handler looks broken when the response is fragmented


lampajr:
**Describe the bug**
If the response is such big that it is fragmented, the JsonParser processor breaks with:

```bash
16:38:59,609 WARN  (epollEventLoopGroup-2-1) [i.h.c.h.j.StreamQueue] Too many buffered fragments, dropping data.
16:38:59,609 WARN  (epollEventLoopGroup-2-1) [i.h.c.h.j.StreamQueue] Too many buffered fragments, dropping data.
16:38:59,609 WARN  (epollEventLoopGroup-2-1) [i.h.c.h.j.StreamQueue] Too many buffered fragments, dropping data.

16:38:59,610 WARN  (epollEventLoopGroup-2-1) [i.h.c.h.j.StreamQueue] Too many buffered fragments, dropping data.
16:38:59,612 ERROR (epollEventLoopGroup-2-1) [i.h.h.s.HttpResponseHandlersImpl] #0 Response body processing failed on io.hyperfoil.http.steps.HttpResponseHandlersImpl@476b0c84 java.lang.IndexOutOfBoundsException: index: -879788, length: 1 (expected: range(0, 65536))
	at io.netty.buffer.AbstractByteBuf.checkRangeBounds(AbstractByteBuf.java:1390)
	at io.netty.buffer.AbstractByteBuf.checkIndex0(AbstractByteBuf.java:1397)
	at io.netty.buffer.AbstractByteBuf.checkIndex(AbstractByteBuf.java:1384)
	at io.netty.buffer.AbstractByteBuf.checkIndex(AbstractByteBuf.java:1379)
	at io.netty.buffer.AbstractByteBuf.getByte(AbstractByteBuf.java:355)
	at io.hyperfoil.core.handlers.json.JsonUnquotingTransformer.transform(JsonUnquotingTransformer.java:48)
	at io.hyperfoil.core.handlers.json.JsonUnquotingTransformer.process(JsonUnquotingTransformer.java:39)
	at io.hyperfoil.core.handlers.json.JsonHandler.record(JsonHandler.java:63)
	at io.hyperfoil.core.handlers.json.StreamQueue.consume(StreamQueue.java:115)
	at io.hyperfoil.core.handlers.json.JsonParser$Context.tryRecord(JsonParser.java:441)
	at io.hyperfoil.core.handlers.json.JsonParser$Context.parse(JsonParser.java:308)
[...]

PR #705, opened 2026-02-24
Add tests for chunked JSON delete transfer and improve JSON handler processing

franz1981:
This is fixing #533 without an aggressive release of buffers once the state machine of JSON parsing progress, since there are cases where it moves backward searching for specific token; if the buffer to read is no longer available, the stream will return -1 making it to believe there is no more data, which is not true.
Retaining the streamed full data is a more conservative fix, while still allowing it to grow to accommodate big chunked JSON.
The original patch to the issue was too aggressive in that, so save memory, but trading with corner cases like these.


franz1981:
@diegolovison hopefully this will solve the regression we observed in the CI introduced by my old patch.

diegolovison (approved):

PR #707, opened 2026-02-25
Add SubMillisecondTimingTest

diegolovison:
Add SubMillisecondTimingTest

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/impl/SimulationRunner.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/SimulationRunner.java
@@ -551,7 +551,7 @@ public class SimulationRunner {
       log.info("GC execution took {} ms", (System.currentTimeMillis() - actualGCRun));
    }
 
-   private void checkNativeTransportForSubMillisecondTiming() {
+   public boolean checkNativeTransportForSubMillisecondTiming() {
       boolean requiresSubMillisecondPrecision = false;
       double threshold = 1000.0;
       for (Phase phase : benchmark.phases()) {
@@ -593,6 +593,7 @@ public class SimulationRunner {
                   "Sub-millisecond rate detected but native epoll transport is unavailable. Timing precision may be limited to millisecond granularity.");
          }
       }
+      return requiresSubMillisecondPrecision;
    }
 
    private static class SharedResources {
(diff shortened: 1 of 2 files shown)


stalep:
## Review

The test intent is good — verifying the sub-millisecond detection logic is valuable. A few concerns:

### Visibility change for testability
Making `checkNativeTransportForSubMillisecondTiming()` from `private` to `public` just for testing leaks an internal implementation detail into `SimulationRunner`'s public API. A cleaner approach would be to extract the detection logic into a static utility method (e.g., in a helper class) that takes a `Benchmark` and `agentId`, making it independently testable without changing `SimulationRunner`'s visibility.

### The fork test documents a potential bug, not correct behavior
`testSubMillisecondDetectionWithForks` asserts `false` for 3000 RPS total split across 4 forks (750 each, below the 1000 threshold). But the system still needs to schedule 3000 events/sec total, which requires sub-millisecond precision. Is this documenting a **bug** (forks should aggregate) or **expected behavior** (each fork has its own `FireTimeSequence` so per-fork rate is what matters)? This should be clarified with a comment.

### Constructor side effects
The test creates a full `SimulationRunner` just to call one method. The constructor likely does significant initialization (event loops, session pools, etc.) that's unnecessary for testing a pure detection method. This further supports extracting the logic into a standalone testable utility.

### Missing `@Tag` annotation
Other standalone benchmark tests use `@Tag("io.hyperfoil.test.Benchmark")`. This test extends `BaseBenchmarkTest` (needs an HTTP server) but doesn't have the tag — it'll run with regular `mvn test` which may or may not be intended.

diegolovison:
@stalep Hello,

* **Visibility change for testability, Constructor side effects**: Not needed
* **The fork test documents a potential bug, not correct behavior**: Can you please rephrase or explain?
* **Missing @Tag annotation**: Done

franz1981:
@diegolovison do we have already other tests which for submilliseconds precision?
Why this one should be different?

We can compute the rate in many different ways, but the way it is honored by the open model is always the same regardless.
I want to understand what this test is reproducing:
- a bug while the ratio is computed to partition the overall 
- a bug on open model while having 1.33 ms interval?


Because the latter is covered and the former doesn't even need to run the workload nor a server, just check which value is passed by computation to the different models.

PR #659, opened 2026-01-20
Add Maven Shade Plugin for wrk and wrk2 CLI executables

diegolovison:
Add Maven Shade Plugin for wrk and wrk2 CLI executables

How to test
```
mvn clean install -DskipTests -pl '!distribution'

java -jar clustering/target/wrk.jar -c 1 -d 1 www.google.com
Running 1 test @ http://www.google.com
  2 threads and 1 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   120.39ms    5.63ms 130.02ms   66.67%
    Req/Sec     9.00      0.00     9.00    100.00
  9 requests in 1.003s, 441.10kB read
Requests/sec: 8.97
Transfer/sec: 439.78kB

java -jar clustering/target/wrk2.jar -c 1 -d 1 --rate 1 www.google.com
Running 1 test @ http://www.google.com
  2 threads and 1 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   116.13ms       0ns 116.39ms    0.00%
    Req/Sec     1.00      0.00     1.00    100.00
  2 requests in 1.003s,  55.16kB read
Requests/sec: 1.99
Transfer/sec:  54.99kB
```

PR #737, opened 2026-03-24
Print total stats summary after run completes in LoadAndRun

stalep:
Addresses jbang-catalog#14 by executing "stats -t" after the wait command completes, so that run@hyperfoil displays the same statistics table (PHASE, METRIC, THROUGHPUT, etc.) that cli@hyperfoil shows.

### DIFF
--- a/cli/src/main/java/io/hyperfoil/cli/Table.java
+++ b/cli/src/main/java/io/hyperfoil/cli/Table.java
@@ -179,7 +179,14 @@ public class Table<T> {
       int suffixLength = prefixes == null ? 0
             : prefixes.stream().filter(Objects::nonNull).mapToInt(Table::width).max().orElse(0);
       int totalWidth = IntStream.of(width).map(w -> w + 2).sum() - 2;
-      int maxWidth = invocation.getShell().size().getWidth() - prefixLength - suffixLength;
+      int shellWidth = invocation.getShell().size().getWidth();
+      if (shellWidth < totalWidth) {
+         int termWidth = org.aesh.readline.util.TerminalUtil.terminalWidth();
+         if (termWidth > shellWidth) {
+            shellWidth = termWidth;
+         }
+      }
+      int maxWidth = shellWidth - prefixLength - suffixLength;
       boolean multiline = totalWidth > maxWidth;
       int idsWidth = IntStream.of(width).limit(idColumns).map(w -> w + 2).sum();
       int stride = width.length - idColumns;
--- a/cli/src/main/java/io/hyperfoil/cli/commands/LoadAndRun.java
+++ b/cli/src/main/java/io/hyperfoil/cli/commands/LoadAndRun.java
@@ -25,7 +25,7 @@ public class LoadAndRun extends BaseStandaloneCommand {
 
    @Override
    protected List<Class<? extends Command<HyperfoilCommandInvocation>>> getDependencyCommands() {
-      return List.of(Upload.class, Wait.class, Report.class);
+      return List.of(Upload.class, Wait.class, Stats.class, Report.class);
    }
 
    @Override
@@ -57,6 +57,7 @@ public class LoadAndRun extends BaseStandaloneCommand {
       @Override
       protected void monitor(HyperfoilCommandInvocation invocation) throws CommandException {
          invocation.executeSwitchable("wait");
+         invocation.executeSwitchable("stats -t");
          if (output != null && !output.isBlank()) {
             invocation.executeSwitchable("report --silent -y --destination " + output);
          } else {


edeandrea:
Hi @stalep do you know when this might be part of a hyperfoil release? We've been waiting on this for some time now (several months).

See quarkusio/spring-quarkus-perf-comparison#42

It would really help us out in some of our investigations.

franz1981:
@edeandrea me or @diegolovison had to roll a new release but we have been busy testing it has no other impacts since we added a lot of changes in the internals 🙏 
Once we were ready to release, the spring vs Quarkus thing happened :)

edeandrea:
> @edeandrea me or @diegolovison had to roll a new release but we have been busy testing it has no other impacts since we added a lot of changes in the internals 🙏 Once we were ready to release, the spring vs Quarkus thing happened :)

Thanks for the update. I think this would be helpful in that investigation, especially being able to quickly test different scenarios without having to update any of the benchmark scripts/infrastructure.

I have https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/113 queued up and ready to go once this is in.

franz1981:
Dis you tried already to see if it looks as you expect @edeandrea ? 🙏

edeandrea:
I can't because this isn't released....

diegolovison:
Now, you can! :)
Take a look at https://github.com/Hyperfoil/Hyperfoil/pull/659

edeandrea:
> Now, you can! :) Take a look at #659

But that means I have to build hyperfoil locally, and then do `java -jar` instead of using hyperfoil via jbang. Thats not a solution unfortunately.

franz1981:
Just be warned @edeandrea :
the result format exposed by Hyperfoil is pretty much stable and is a stronger contract that the cli output.
For wrk, nope because wrk was based on an existing tool which hyperfoil has to emulate, output included

edeandrea:
> Just be warned @edeandrea : the result format exposed by Hyperfoil is pretty much stable and is a stronger contract that the cli output. For wrk, nope because wrk was based on an existing tool which huperfoil has to emulate, output included

But maybe i'm mistaken, but when I run hyperfoil via jbang it does not output anywhere the req/sec number. You'd have to go searching through lots of files to find numbers and then do some math.

I'm not saying that the cli output needs to match wrk character for character, but when I run hyperfoil manually it outputs this:


```
[hyperfoil@in-vm]$ run app-load-test
Started run 0010
Total stats from run 0010
PHASE     METRIC         THROUGHPUT     REQUESTS  MEAN     STD_DEV  MAX       p50      p90      p99       p99.9     p99.99    TIMEOUTS  ERRORS  BLOCKED  2xx      3xx  4xx  5xx  CACHE
warmup    executeAction  8862.38 req/s   1063618  5.64 ms  2.84 ms  55.05 ms  4.95 ms  8.98 ms  16.12 ms  27.66 ms  39.32 ms         0       0     0 ns  1063618    0    0    0      0
loadTest  executeAction  8728.55 req/s    261900  5.73 ms  2.87 ms  55.57 ms  5.01 ms  9.18 ms  16.45 ms  26.87 ms  37.75 ms         0       0     0 ns   261900    0    0    0      0
```

But when I run hyperfoil via jbang, it doesn't print any of that. Thats the data I'm after with this change

diegolovison:
- Build hyperfoil locally
- Copy the jbang command
- Replace with java -jar
- See if it matches your expectations.

You don't need a SUT. Do it super simple like `-c 1 -d 1 www.google.com`

edeandrea:
> * Build hyperfoil locally
> * Copy the jbang command
> * Replace with java -jar
> * See if it matches your expectations.
> 
> You don't need a SUT. Do it super simple like `-c 1 -d 1 www.google.com`

Are you saying do that just to ensure the change works as I expect it?

franz1981:
Yes but you will parse that data right?
What I know is that the cli output is not a strong contract like the result file and afaik all the other tests built by my team just extract what's needed from there computing it themselves.
That said I understand is not as convenient as while making the cli to compute it but allows you to be in control and don't break - silently - if we add a new column (which has happened ..).

In addition to that, I remember we talked already about this feature and as long as we don't run fixed TPS in the quarkus benchmark, having the html files either is just misleading.
In all out TPS, the load generator act in closed model model, and the histograms of latencies are prone to coordinated omission and misleading..

The sole case where it would be convenient is by having multiple endpoints

franz1981:
@edeandrea the release has been done and the jbang catalog updated. I have to check myself to see if it works, but is there :+1:

edeandrea:
It works!

edeandrea:
Is there a way I can "pin" to a specific version? Both `run@hyperfoil` and `cli@hyperfoil` both have the same `script-ref`.

If I use `jbang run@hyperfoil`, the version of hyperfoil is floating. In my case I want to pin to a specific hyperfoil version.

PR #769, opened 2026-04-29
Bump vertx to 5.0.12

diegolovison:
**Bump vertx to 5.0.12**

Dependency upgrade:
- infinispan: 14.0.25
- jgroups: 5.2.22


diegolovison:
@jabolina can you please review the `infinispan.xml` and `jgroups-tcp-*.xml`

jabolina (approved):
Looks good.

diegolovison:
@franz1981 please review

franz1981:
One behavioral change to address:

 - persistRun (ControllerVerticle.java): master used future.fail(e) after CSV write failure, which logged the error but continued executing subsequent persistence steps (info.json, stats.json, hooks). The new throw new RuntimeException(e) aborts on first error — downstream steps (especially post-hooks) won't run. Consider catching per-step and collecting errors, or wrapping each step individually.

Minor: new DatabindCodec() in RunRefImpl creates a private ObjectMapper without registered subtypes — works for current targets but fragile if deserialization targets grow.

clustering/src/main/java/io/hyperfoil/clustering/OutputStreamAdapter.java
```diff
@@ -15,12 +14,12 @@ public OutputStreamAdapter(WriteStream<Buffer> stream) {
 
    @Override
    public void write(byte[] b) {
-      stream.write(Buffer.buffer(Unpooled.wrappedBuffer(b)));
+      stream.write(Buffer.buffer(b));
```
stalep: **Performance regression:** `Buffer.buffer(b)` copies the byte array (allocates a new heap ByteBuf via `writeBytes(bytes)`). The old `Unpooled.wrappedBuffer(b)` was zero-copy -- it wrapped the array directly as the ByteBuf backing store.

This class streams data through a `WriteStream<Buffer>`, so for large payloads (benchmark results, statistics) this means unnecessary allocation + copy.

The old code still works in Vert.x 5 since `Buffer.buffer(ByteBuf)` still wraps without copying:

```java
stream.write(Buffer.buffer(Unpooled.wrappedBuffer(b)));
```

Same issue on line 22 with `Buffer.buffer().appendBytes(b, off, len)` -- should remain:
```java
stream.write(Buffer.buffer(Unpooled.wrappedBuffer(b, off, len)));
```

clustering/src/main/resources/jgroups-tcp.xml
```diff
@@ -43,14 +45,12 @@
     <pbcast.STABLE desired_avg_gossip="5000"
                    max_bytes="1M"
     />
-    <pbcast.GMS print_local_addr="false"
-                join_timeout="${jgroups.join_timeout:2000}"
-    />
+    <pbcast.GMS print_local_addr="false" join_timeout="0"/>
```
stalep, line 48: **Likely a bug:** `join_timeout` changed from `${jgroups.join_timeout:2000}` to hardcoded `"0"`.

- The agent config (`jgroups-tcp-agent.xml`) still uses `${jgroups.join_timeout:2000}` -- asymmetry.
- Infinispan 14's own default TCP config uses `join_timeout="${jgroups.join_timeout:2000}"`.
- JGroups 5.2.x `conf/tcp.xml` uses `join_timeout="2000"`.
- `join_timeout=0` means the controller won't wait for join responses at all.

Was this intentional? If so, it removes the ability to configure via system property. Recommend keeping `${jgroups.join_timeout:2000}` for consistency.

http/src/main/java/io/hyperfoil/http/connection/HttpClientPoolImpl.java
```diff
@@ -194,6 +194,15 @@ private SslContext createSslContext() throws SSLException {
...
             .keyManager(createKeyManagerFactory());
+
+      // For Vert.x 5 compatibility: Netty now enforces stricter hostname verification by default.
+      // Disable endpoint identification when using InsecureTrustManagerFactory or when the system
+      // property is set (for testing with self-signed certificates that don't have proper SANs)
+      if (trustManagerFactory == InsecureTrustManagerFactory.INSTANCE ||
+            Boolean.getBoolean("io.hyperfoil.http.disableEndpointIdentification")) {
+         builder.endpointIdentificationAlgorithm(null);
+      }
+
```
stalep, line 206: The `io.hyperfoil.http.disableEndpointIdentification` property should follow the project convention and be declared as a constant in `Properties.java`, then accessed via `Properties.getBoolean()`. Two reasons:

1. **Discoverability** -- all other system properties are centralized in `Properties.java`.
2. **Env var support** -- `Properties.getBoolean()` also checks environment variables (uppercased with underscores). `Boolean.getBoolean()` only checks system properties.

Also consider a dot-separated property name matching the `io.hyperfoil.*` conventions.

http/src/main/java/io/hyperfoil/http/connection/HttpClientPoolImpl.java
```diff
@@ -338,6 +347,8 @@ void connect(final HttpConnectionPool pool, ConnectionReceiver handler) {
       bootstrap.group(pool.executor());
       bootstrap.option(ChannelOption.SO_KEEPALIVE, true);
       bootstrap.option(ChannelOption.SO_REUSEADDR, true);
+      // Use pooled allocator for Vert.x 5 compatibility
+      bootstrap.option(ChannelOption.ALLOCATOR, io.netty.buffer.PooledByteBufAllocator.DEFAULT);
```
stalep: The comment says "Use pooled allocator for Vert.x 5 compatibility" but `PooledByteBufAllocator.DEFAULT` has been Netty's default allocator since Netty 4.1, so this should be a no-op. Was there a specific test failure that required this? If so, the comment should describe the observed issue. If not, this line can be removed.

http/src/test/java/io/hyperfoil/http/CertificatesTest.java
```diff
@@ -34,6 +36,13 @@
 
 @ExtendWith(VertxExtension.class)
 public class CertificatesTest {
+
+   @BeforeAll
+   public static void setupPooledBuffers() {
+      // Disable endpoint identification for tests with self-signed certificates
+      System.setProperty("io.hyperfoil.http.disableEndpointIdentification", "true");
+   }
```
stalep, line 51: Two issues:

1. **System property leak:** Set in `@BeforeAll` but never cleared in `@AfterAll`. If tests run in the same JVM (Surefire default), this leaks to other test classes and could mask real SSL bugs.

2. **Method name:** `setupPooledBuffers` sets an endpoint identification property, not anything related to pooled buffers.

Suggested fix:
```java
@BeforeAll
public static void disableEndpointIdentification() {
   System.setProperty("io.hyperfoil.http.disableEndpointIdentification", "true");
}

@AfterAll
public static void restoreEndpointIdentification() {
   System.clearProperty("io.hyperfoil.http.disableEndpointIdentification");
}
```

pom.xml
```diff
@@ -88,13 +88,18 @@
...
         <version.slf4j>2.0.18</version.slf4j>
         <version.snakeyaml>2.6</version.snakeyaml>
-        <version.vertx>4.5.26</version.vertx>
+        <version.vertx>5.0.12</version.vertx>
+
         <!-- version.infinispan must match what came from vertx-infinispan version -->
-        <version.infinispan>13.0.17.Final</version.infinispan>
+        <version.infinispan>14.0.25.Final</version.infinispan>
         <version.caffeine>2.8.4</version.caffeine>
         <version.rxjava>3.0.4</version.rxjava>
```
stalep: **RxJava version downgrade:** This pins `3.0.4` but `vertx-infinispan` 5.0.12 declares a dependency on RxJava `3.0.13`. Hyperfoil's `dependencyManagement` will override the transitive version, downgrading to `3.0.4`. This could cause compatibility issues if `vertx-infinispan` relies on fixes in 3.0.5-3.0.13.

Recommend bumping to at least `3.0.13` to match what `vertx-infinispan` expects.

clustering/src/main/resources/jgroups-tcp.xml
```diff
@@ -3,18 +3,20 @@
         xsi:schemaLocation="urn:org:jgroups http://www.jgroups.org/schema/jgroups-4.0.xsd">
```
stalep: The JGroups XSD schema reference is still `jgroups-4.0.xsd` but this PR upgrades to JGroups 5.2.x (via Infinispan 14). Should update to `jgroups-5.2.xsd`:

```xml
xsi:schemaLocation="urn:org:jgroups http://www.jgroups.org/schema/jgroups-5.2.xsd"
```

Same applies to `jgroups-tcp-agent.xml`.

http/src/test/java/io/hyperfoil/http/MemoryUsageTest.java
```diff
@@ -77,11 +78,13 @@ protected void test(VertxTestContext context, HttpServerOptions serverOptions) {
             context.failNow(e);
          }
       };
-      Vertx.vertx().createHttpServer(serverOptions)
+      // Create Vertx with pooled buffers enabled for Vert.x 5
+      VertxOptions vertxOptions = new VertxOptions();
+      Vertx.builder().with(vertxOptions).build().createHttpServer(serverOptions)
             .requestHandler(ctx -> ctx.response()
```
stalep: The comment says "Create Vertx with pooled buffers enabled for Vert.x 5" but the `VertxOptions` is empty -- no pooled buffer option is set. `Vertx.builder().with(new VertxOptions()).build()` is functionally identical to `Vertx.vertx()`. Either set the relevant option or simplify back to `Vertx.vertx()`.

stalep (changes requested):
Solid work on this major upgrade - the mechanical Vert.x 5 migrations (listen/send/executeBlocking patterns, Promise no longer extending Handler, CompositeFuture removal, RoutingContext body API) are all correct. The AgentVerticle deployment ID rewrite for cluster-wide uniqueness is a good catch.

Franz already flagged the `persistRun` behavioral change and the `DatabindCodec` subtype issue, so I won't repeat those. Here are additional items.

### Behavioral change in `ControllerVerticle.java:494`

The condition for calling `failNotStartedPhases` changed from:
```java
if (controllerPhase.isFailed()) {
```
to:
```java
if (controllerPhase.isFailed() && run.benchmark.failurePolicy() == Benchmark.FailurePolicy.CANCEL) {
```

This means phases are no longer cancelled on failure unless the benchmark explicitly sets a `CANCEL` policy. Was this intentional? What is the default `failurePolicy()` value, and what should happen when a phase fails without an explicit policy?

franz1981:
And @diegolovison IIRC Vertx 5 has the option to provide SPI which can capture the event loop group and install what we need -> this is likely necessary to allow infinispan to work reliably...

clustering/src/main/resources/jgroups-tcp.xml
diegolovison: @jabolina you set it to zero at https://github.com/Hyperfoil/Hyperfoil/commit/bc123e32193460c17edc071f5ccc7b5adea30d7b#diff-6112155d22709b221b4749c8a899802264945002a74f73e0d107431e73fdfaaeR50

What's your opinion?

http/src/main/java/io/hyperfoil/http/connection/HttpClientPoolImpl.java
diegolovison: There will be erors in the test
``` 
[ERROR] Errors: 
[ERROR]   MemoryUsageTest.testEncryptHttp1x(VertxTestContext) � Vertx Buffers are not pooled
[ERROR]   MemoryUsageTest.testEncryptHttp2(VertxTestContext) � Vertx Buffers are not pooled
[ERROR]   MemoryUsageTest.testPlainHttp1x(VertxTestContext) � Vertx Buffers are not pooled
[INFO] 
[ERROR] Tests run: 134, Failures: 0, Errors: 3, Skipped: 0
```

clustering/src/main/resources/jgroups-tcp.xml
jabolina: IIRC, it was because of how we deployed the controller and agents. It was supposed to always be a single controller first, and the agents later, to run the benchmark.

Since we always start with a single controller, there was no need to wait for the join, because it was always going to be only this single node. The agents have a timeout because they could be deployed concurrently and need to join with the controller.

If Hyperfoil deploys differently, i.e., multiple controllers, then it makes sense to have an initial timeout. Having it configurable wouldn't be a bad thing, if ever needed to override it.

http/src/main/java/io/hyperfoil/http/connection/HttpClientPoolImpl.java
franz1981: One second, why we do't use the adaptive from Netty  4.2?

What the test...is testing?
And under which scenario?

In theory Vertx should use "pooled" buffers regardless 

I'm checking this...

franz1981: @diegolovison 

AdaptiveByteBufAllocator has usedDirectMemory() and implements ByteBufAllocatorMetricProvider. 
So the test could work with the adaptive allocator — just needs to check instanceof ByteBufAllocatorMetricProvider instead of instanceof PooledByteBufAllocator.

TLDR bad test ^^

franz1981: FYI 
```java
● Update(http/src/test/java/io/hyperfoil/http/MemoryUsageTest.java)
  ⎿  Added 2 lines, removed 2 lines
      109                 ByteBufAllocator alloc = data.alloc();
      110                 if (!data.isDirect()) {
      111                    context.failNow("Expecting to use direct buffers");
      112 -               } else if (alloc instanceof PooledByteBufAllocator) {                                                                                  
      113 -                  long usedMemory = ((PooledByteBufAllocator) alloc).metric().usedDirectMemory();                                                     
      112 +               } else if (alloc instanceof ByteBufAllocatorMetricProvider) {                                                                          
      113 +                  long usedMemory = ((ByteBufAllocatorMetricProvider) alloc).metric().usedDirectMemory();                                             
      114                    if (usedMemory < 0) {
      115                       context.failNow("Cannot fetch direct memory stats");
      116                    }
```

diegolovison:
@franz1981 @stalep thanks for your valuable review.
There are some pending items, but others were addressed per commit.
Before merging, I wish to squash.



- [x] persistRun (ControllerVerticle.java): master used future.fail(e) after CSV write failure, which logged the error but continued executing subsequent persistence steps (info.json, stats.json, hooks). The new throw new RuntimeException(e) aborts on first error — downstream steps (especially post-hooks) won't run. Consider catching per-step and collecting errors, or wrapping each step individually.

- [x] Minor: new DatabindCodec() in RunRefImpl creates a private ObjectMapper without registered subtypes — works for current targets but fragile if deserialization targets grow.

- [x] Performance regression: Buffer.buffer(b) copies the byte array (allocates a new heap ByteBuf via writeBytes(bytes)). The old Unpooled.wrappedBuffer(b) was zero-copy -- it wrapped the array directly as the ByteBuf backing store.

- [ ] Likely a bug: join_timeout changed from ${jgroups.join_timeout:2000} to hardcoded "0".

- [x] The io.hyperfoil.http.disableEndpointIdentification property should follow the project convention and be declared as a constant in Properties.java, then accessed via Properties.getBoolean(). Two reasons:

- [x] "Use pooled allocator for Vert.x 5 compatibility" but PooledByteBufAllocator.DEFAULT has been Netty's default allocator since Netty 4.1

- [x] Restore Endpoint Identification

- [x] vertx-infinispan 5.0.12 declares a dependency on RxJava 3.0.13

- [x] Set JGroups XSD schema to 5.2

- [x] Removed "Create Vertx with pooled buffers enabled for Vert.x 5" comments
----

> This means phases are no longer cancelled on failure unless the benchmark explicitly sets a CANCEL policy. Was this intentional? What is the default failurePolicy() value, and what should happen when a phase fails without an explicit policy?

I was done at https://github.com/Hyperfoil/Hyperfoil/commit/1ed392846bd11b0d5aca77eb4a95d8d78c15e2df 
It is not related to this PR

clustering/src/main/resources/jgroups-tcp.xml
diegolovison: @stalep, any objection?

http/src/main/java/io/hyperfoil/http/connection/HttpClientPoolImpl.java
diegolovison: Done

diegolovison:
@franz1981 @stalep would like to review? The review is separated by commit

stalep (approved):
All review feedback has been addressed well -- nice work on the follow-up commits.

Quick summary of verifications:
- **persistRun**: best-effort behavior restored with `hasError` flag -- all steps run, throws at the end.
- **DatabindCodec**: uses `DatabindCodec.mapper()` via `decode()` helper -- subtypes will resolve correctly.
- **OutputStreamAdapter**: zero-copy restored with `Unpooled.wrappedBuffer()` + `BufferInternal.buffer()`.
- **Properties.java**: `DISABLE_ENDPOINT_IDENTIFICATION` added with dot-separated naming, accessed via `Properties.getBoolean()`.
- **CertificatesTest**: `@AfterAll` cleanup added, uses the constant.
- **RxJava**: bumped to 3.0.13, duplicate declaration removed.
- **JGroups XSD**: both files updated to 5.2.
- **PooledByteBufAllocator**: removed. `MemoryUsageTest` now uses `ByteBufAllocatorMetricProvider` interface instead -- good fix.
- **Misleading comments**: removed.

One remaining item per your own checklist: **`join_timeout="0"`** in `jgroups-tcp.xml` is still hardcoded. Should be resolved before merge.

I would like @franz1981 to have a look as well though.

diegolovison:
> One remaining item per your own checklist: join_timeout="0" in jgroups-tcp.xml is still hardcoded. Should be resolved before merge.

@stalep I think we can keep it as 0 because there is only one controller

franz1981 (approved):
Well done 💯

PR #770, opened 2026-05-04
Replace SCP with SFTP for agent file deployment

diegolovison:
Problem:
The original problem was: ClusterTestCase.startClusteredBenchmarkTest() was failing with java.io.EOFException: readAck - EOF before ACK when deploying agents with 102 JAR files each. The SSH server was closing connections after ~100 SCP file transfers.

Fix: Replace SCP with SFTP for agent file deployment

### DIFF
--- a/clustering/src/main/java/io/hyperfoil/deploy/ssh/SshDeployedAgent.java
+++ b/clustering/src/main/java/io/hyperfoil/deploy/ssh/SshDeployedAgent.java
@@ -2,8 +2,10 @@ package io.hyperfoil.deploy.ssh;
 
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.PrintStream;
@@ -23,8 +25,6 @@ import org.apache.sshd.common.io.IoOutputStream;
 import org.apache.sshd.common.io.IoReadFuture;
 import org.apache.sshd.common.util.buffer.ByteArrayBuffer;
 import org.apache.sshd.common.util.io.output.NullOutputStream;
-import org.apache.sshd.scp.client.ScpClient;
-import org.apache.sshd.scp.client.ScpClientCreator;
 import org.apache.sshd.sftp.client.SftpClient;
 import org.apache.sshd.sftp.client.SftpClientFactory;
 
@@ -56,7 +56,7 @@ public class SshDeployedAgent implements DeployedAgent {
    private ClientSession session;
    private ChannelShell shellChannel;
    private Consumer<Throwable> exceptionHandler;
-   private ScpClient scpClient;
+   private SftpClient sftpClient;
    private PrintStream commandStream;
 
    public SshDeployedAgent(String name, String runId, String username, String hostname, String sshKey, int port, String dir,
@@ -79,6 +79,13 @@ public class SshDeployedAgent implements DeployedAgent {
          commandStream.close();
       }
       try {
+         if (sftpClient != null) {
+            sftpClient.close();
+         }
+      } catch (IOException e) {
+         log.error("Failed closing SFTP client", e);
+      }
+      try {
          if (shellChannel != null) {
             shellChannel.close();
          }
@@ -96,7 +103,13 @@ public class SshDeployedAgent implements DeployedAgent {
       this.session = session;
       this.exceptionHandler = exceptionHandler;
 
-      this.scpClient = ScpClientCreator.instance().createScpClient(session);
+      // Create SFTP client with the session - SFTP is more efficient for multiple file transfers
+      try {
+         this.sftpClient = SftpClientFactory.instance().createSftpClient(session);
+      } catch (IOException e) {
+         exceptionHandler.accept(new DeploymentException("Failed to create SFTP client", e));
+         return;
+      }
 
       try {
          this.shellChannel = session.createShellChannel();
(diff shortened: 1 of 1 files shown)


diegolovison:
@stalep @franz1981 the only way to fix the issue on my machine was to change to SFTP. That, imho, is the right way to transfer files.

franz1981:
But which version introduced this bug and why?

diegolovison:
It is something with my computer. Then, using the proper method (SFTP) fixed it for me.
I think this is how it should be since the beginning.

clustering/src/main/java/io/hyperfoil/deploy/ssh/SshDeployedAgent.java
```diff
@@ -231,6 +259,17 @@ public void close() throws IOException {
                   "Agent process terminated prematurely. Hint: type 'log " + name + "' to see agent output.")));
    }
 
+   private void uploadViaSftp(String localPath, String remotePath) throws IOException {
+      try (InputStream in = new FileInputStream(localPath);
+            OutputStream out = sftpClient.write(remotePath)) {
+         byte[] buffer = new byte[8192];
```
franz1981, line 270: This doesn't look great. how scp does it right now?

franz1981 (changes requested):

clustering/src/main/java/io/hyperfoil/deploy/ssh/SshDeployedAgent.java
```diff
@@ -96,7 +103,13 @@ public void deploy(ClientSession session, Consumer<Throwable> exceptionHandler)
...
       this.exceptionHandler = exceptionHandler;
 
-      this.scpClient = ScpClientCreator.instance().createScpClient(session);
+      // Create SFTP client with the session - SFTP is more efficient for multiple file transfers
+      try {
+         this.sftpClient = SftpClientFactory.instance().createSftpClient(session);
+      } catch (IOException e) {
+         exceptionHandler.accept(new DeploymentException("Failed to create SFTP client", e));
+         return;
+      }
```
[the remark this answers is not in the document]
diegolovison: Done

clustering/src/main/java/io/hyperfoil/deploy/ssh/SshDeployedAgent.java
diegolovison: The current SFTP implementation uses a standard 8KB buffer with try-with-resources for proper cleanup. 

Are you concerned about:
1. Performance (buffer size, transfer speed)?
2. Resource management (connection handling)?
3. Error handling?

I can adjust the implementation based on your feedback.

diegolovison:
@franz1981 anything else?

clustering/src/main/java/io/hyperfoil/deploy/ssh/SshDeployedAgent.java
franz1981: What the existing SCP impl was doing instead? 
What are the pros/cons between the twos if there is lot to be exchange over the wire?

diegolovison: See the comparison. The test is attached to this message.

```
=== SCP vs SFTP Performance Comparison ===
Host: dlovison@localhost:22
Remote directory: /tmp/hyperfoil-ssh-test
Seed: 42

--- Testing with 1 file(s) ---
Total size: 0.51 MB
SCP:     304 ms (1.67 MB/s)
SFTP:     16 ms (31.72 MB/s)
SFTP is 94.7% faster than SCP

--- Testing with 10 file(s) ---
Total size: 5.96 MB
SCP:    2793 ms (2.13 MB/s)
SFTP:     44 ms (135.52 MB/s)
SFTP is 98.4% faster than SCP

--- Testing with 100 file(s) ---
Total size: 49.15 MB
SCP:   29255 ms (1.68 MB/s)
SFTP:    349 ms (140.82 MB/s)
SFTP is 98.8% faster than SCP

--- Testing with 1000 file(s) ---
Total size: 495.83 MB
SCP:  363706 ms (1.36 MB/s)
SFTP:   4894 ms (101.31 MB/s)
SFTP is 98.7% faster than SCP
```
[ScpVsSftpPerformanceTest.java](https://github.com/user-attachments/files/27489034/ScpVsSftpPerformanceTest.java)

diegolovison: **SCP**
Pros:
- Preserves file attributes automatically

Cons:
- Less efficient for multiple files - creates new connection/channel per transfer
- Higher protocol overhead
- No connection reuse between transfers
- More resource intensive for batch operations

**SFTP**
Pros:
- Much more efficient for multiple files - reuses single persistent connection
- Lower protocol overhead per file
- Better performance for batch transfers (as noted in code comment: "SFTP is more efficient for multiple file transfers")
- More control over transfer process
- Better suited for modern SSH implementations

Cons:
- Doesn't preserve file attributes by default
- Requires explicit resource management

franz1981: why it is faster? did you profiled it? 

And..it is the hyperfoil same code? or a separate test?

diegolovison: Yeah. The test has the same code that Hyperfoil had/have.
Profiled using time elapsed the discrepancy is huge >90%

diegolovison: image
The code is attached to the comment

franz1981: My point is to find what was the problem with scp because maybe that can fix your case too

diegolovison:
Closed in favor https://github.com/Hyperfoil/Hyperfoil/pull/778

PR #778, opened 2026-05-08
Add configurable file transfer client with SCP/SFTP support

diegolovison:
Add configurable file transfer client with SCP/SFTP support

### DIFF
--- a/api/src/main/java/io/hyperfoil/internal/Properties.java
+++ b/api/src/main/java/io/hyperfoil/internal/Properties.java
@@ -41,6 +41,7 @@ public interface Properties {
    String GC_CHECK = "io.hyperfoil.gc.check.enabled";
    String CLUSTER_JGROUPS_STACK = "io.hyperfoil.cluster.jgroups_stack";
    String REPORT_TEMPLATE = "io.hyperfoil.report.template";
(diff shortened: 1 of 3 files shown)


diegolovison:
@franz1981, this will allow us to troubleshoot better the scp VS sftp. Also, it keeps by the default scp, and it will unblock my work because I can use `mvn package -Pbenchmark -pl test-suite -am -Dtest=ClusterTestCase -Dsurefire.failIfNoSpecifiedTests=false -Dio.hyperfoil.file.transfer.client=sftp`

WDYT?

franz1981:
To me is fine, but after we optimized scp to use a single connection, is it that terrible?

Asking because maybe you want to push a change to enable the optimized version of scp to be used too

diegolovison:
@franz1981 
- Created a new commit to make it easy to review the code. We can squash before the merge
- Created https://github.com/Hyperfoil/Hyperfoil/issues/779 to profile scp and sftp when it gets the priority.

franz1981:
Why you need to profile now if we understood that the problem is the number of connections? 🙏

diegolovison:
Closed in favor of https://github.com/Hyperfoil/Hyperfoil/pull/780

PR #780, opened 2026-05-11
Optimize SSH file transfer by batching uploads

diegolovison:
Optimize SSH file transfer by batching uploads

This can boost the copy by around 10x
Here is the raw data and ratio.

| Files | scp (Baseline) | scp with array |
|---|---|---|
| 1 | 336 ms (1.0x) | 292 ms (1.15x faster) |
| 10 | 3235 ms (1.0x) | 533 ms (6.07x faster) |
| 100 | 33482 ms (1.0x) | 3243 ms (10.32x faster) |
| 1000 | 319168 ms (1.0x) | 32844 ms (9.72x faster) |

Reproducer. You need to remove the files because there is a checksum in the code to prevent copying again.
``` 
rm -fR /home/dlovison/.m2/repository/io/hyperfoil/hyperfoil-clustering
rm -fR /tmp/hyperfoil/
mvn clean package -Pbenchmark
```

### DIFF
--- a/clustering/src/main/java/io/hyperfoil/deploy/ssh/SshDeployedAgent.java
+++ b/clustering/src/main/java/io/hyperfoil/deploy/ssh/SshDeployedAgent.java
@@ -8,7 +8,9 @@ import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
@@ -166,20 +168,25 @@ public class SshDeployedAgent implements DeployedAgent {
       String java = Properties.get(Properties.AGENT_JAVA_EXECUTABLE, "java");
       startAgentCommmand.append(java).append(" -cp ");
 
+      List<String> fileToUpload = new ArrayList<>();
       for (Map.Entry<String, String> entry : localMd5.entrySet()) {
          int lastSlash = entry.getKey().lastIndexOf("/");
          String filename = lastSlash < 0 ? entry.getKey() : entry.getKey().substring(lastSlash + 1);
          String remoteChecksum = remoteMd5.remove(filename);
          if (!entry.getValue().equals(remoteChecksum)) {
             log.debug("MD5 mismatch {}/{}, copying {}", entry.getValue(), remoteChecksum, entry.getKey());
-            try {
-               scpClient.upload(entry.getKey(), dir + AGENTLIB + "/" + filename, ScpClient.Option.PreserveAttributes);
-            } catch (IOException e) {
-               exceptionHandler.accept(e);
-               return;
-            }
+            fileToUpload.add(entry.getKey());
          }
          startAgentCommmand.append(dir).append(AGENTLIB).append('/').append(filename).append(':');
+      }
+      if (!fileToUpload.isEmpty()) {
+         try {
+            scpClient.upload(fileToUpload.toArray(String[]::new), dir + AGENTLIB + "/",
+                  ScpClient.Option.PreserveAttributes);
+         } catch (IOException e) {
+            exceptionHandler.accept(e);
+            return;
+         }
       }
       if (!remoteMd5.isEmpty()) {
          StringBuilder rmCommand = new StringBuilder();
--- a/test-suite/src/test/java/io/hyperfoil/benchmark/BaseWrkBenchmarkTest.java
+++ b/test-suite/src/test/java/io/hyperfoil/benchmark/BaseWrkBenchmarkTest.java
@@ -11,7 +11,7 @@ import io.vertx.ext.web.Router;
  */
 public abstract class BaseWrkBenchmarkTest extends BaseBenchmarkTest {
 
-   private final long unservedDelay = 2000;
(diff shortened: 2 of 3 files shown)


franz1981:
> 1000 files: 89%

1000 files is not just 89% faster than the existing approach 🙏 

Use the ratio as a measure of improvement

diegolovison:
Here is the raw data and ratio.

| Files | scp (Baseline) | scp with array |
|---|---|---|
| 1 | 336 ms (1.0x) | 292 ms (1.15x faster) |
| 10 | 3235 ms (1.0x) | 533 ms (6.07x faster) |
| 100 | 33482 ms (1.0x) | 3243 ms (10.32x faster) |
| 1000 | 319168 ms (1.0x) | 32844 ms (9.72x faster) |

diegolovison:
@franz1981 CI was complaining. We were unlucky, and then I reduced the `unservedDelay`
This is ready to be merged.

clustering/src/main/java/io/hyperfoil/deploy/ssh/SshDeployedAgent.java
```diff
@@ -166,21 +168,26 @@ public void close() throws IOException {
...
-               exceptionHandler.accept(e);
-               return;
-            }
+            fileToUpload.add(entry.getKey());
          }
          startAgentCommmand.append(dir).append(AGENTLIB).append('/').append(filename).append(':');
       }
+      if (!fileToUpload.isEmpty()) {
+         try {
+            scpClient.upload(fileToUpload.toArray(String[]::new), dir + AGENTLIB + "/",
```
franz1981, line 184: maybe we want to have a configurable amount of connections?
The scp changes are now using a single connection instead of 1 per file, but maybe we can have 10 connections for 100 files etc etc no?

The default can still 1 but people can change it, with a max == total number of files (so you can have the default behaviour with -1 or 0 - what fit best)

franz1981 (changes requested):

diegolovison:
> maybe we want to have a configurable amount of connections?

I think we can add if really needed. Otherwise, it is one more scenario to maintain.

franz1981:
I don't want to assume 1 connection fit all yet: you tried on your machine only so far no?
If users complain we need a way to make them switch back to the original version without changing hyperfoil version

clustering/src/main/java/io/hyperfoil/deploy/ssh/SshDeployedAgent.java
diegolovison: Can it be a percentage?
Because the total number of files is going to change from release to release.

franz1981:
Usually you want to setup the number of physical resources required; a percentage is dangerous as it can grow a lot...while if you tell people to setup 1000 they probably think twice than 100% without knowing that hyperfoil jars can be 200..

diegolovison:
Ok. Let's use the number of connections. 

Example: 
- 1: All files copied once
- 2: Divide the number of files by 2
- 3: Divide the number of files by 3

Should 1 be the default?

franz1981:
> Should 1 be the default?

only if we can check on the CI how it behaves...

diegolovison:
How about keeping by default a connection per file (no extra parameter), and if someone sets the parameter `-Dio.hyperfoil.file.transfer.client.scp.connections=1`, then:
1: All files copied once
2: Divide the number of files by 2
3: Divide the number of files by 3

franz1981:
> How about keeping by default a connection per file 

that's the current behavior you mean?

diegolovison:
The current behavior is that all files are in a single connection. 

---

My proposal is to change the default behavior to keep it as it was before this PR
- Iterate over each file and call `upload`

----

Then, introduce a new parameter.
If we set `-Dio.hyperfoil.file.transfer.client.scp.connections=1` then, we have a call to `upload` with all files

franz1981:
> The current behavior is that all files are in a single connection.

TBH I was thinking it was creating a new connection per each file - instead. Batching on the other side batch ALL files to a single scp connection

franz1981:
@diegolovison 

The current code (before your PR) does not put all files in a single connection. It calls scpClient.upload(file, remote) inside a loop — and each call is blocking: it opens a new SCP channel, transfers the file, waits for the ACK, and closes the channel before the next iteration starts.

So with 1000 files, you get 1000 sequential channel-open/transfer/close cycles.

Your PR collects all files into a list and calls scpClient.upload(files[], remote) once — one channel, all files streamed through it back-to-back, no repeated handshake overhead.  That's where the 10x is coming from, likely (to verify it!).

My suggestion at this point is NOT to make it configurable, as the old behavior was just wrong.
You can verify it was using blocking SCP calls yourself please :pray:

diegolovison:
@franz1981, what you describe is accurate.

> My suggestion at this point is NOT to make it configurable

That is good! Ready to merge?

franz1981 (approved):
Well done!

franz1981:
But check if there are hidden timeout on scp now that the single connection run for longer please 🙏

diegolovison:
I don't see any timeout, and all tests are passing. If there are timeouts ( like before this change ), the test will fail ( like it was on my computer )

franz1981:
> I don't see any timeout

I mean; check the scp wrapper library default timeout as we need to make sure they want fall shortly based on the available bandwidth. We need them to be enlarged to accomodate to the larger quantity of files sent

PR #784, opened 2026-05-12
Add configurable file transfer client with SCP/SFTP support

diegolovison:
Add a configurable file transfer client with SCP/SFTP support

The default behavior will be scp with array, but the user is allowed to change to sftp via `-Dio.hyperfoil.file.transfer.client=sftp`

This can boost the copy by around 11x

Here is the raw data and ratio.
| Files | scp with array (Baseline) | sftp  |
|---|---|---|
| 1 | 304 ms | 15 ms (20.27x faster) |
| 10 | 540 ms | 50 ms (10.80x faster) |
| 100 | 3442 ms | 296 ms (11.63x faster) |
| 1000 | 33474 ms | 2760 ms (12.13x faster) |

Reproducer:
```
mvn test -Dtest=FileTransferPerformanceTest -Dio.hyperfoil.file.transfer.client=scp
mvn test -Dtest=FileTransferPerformanceTest -Dio.hyperfoil.file.transfer.client=sftp
```

### DIFF
--- a/clustering/src/main/java/io/hyperfoil/deploy/ssh/FileTransfer.java
+++ b/clustering/src/main/java/io/hyperfoil/deploy/ssh/FileTransfer.java
@@ -1,0 +1,110 @@
+package io.hyperfoil.deploy.ssh;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.List;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.sshd.client.session.ClientSession;
+import org.apache.sshd.scp.client.ScpClient;
+import org.apache.sshd.scp.client.ScpClientCreator;
+import org.apache.sshd.sftp.client.SftpClient;
+import org.apache.sshd.sftp.client.SftpClientFactory;
+
+import io.hyperfoil.internal.Properties;
+
+public class FileTransfer implements AutoCloseable {
+
+   private static final Logger log = LogManager.getLogger(FileTransfer.class);
+
+   private IFileTransfer client;
+
+   FileTransfer(ClientSession session) throws Exception {
+      String fileTransferClient = Properties.get(Properties.FILE_TRANSFER_CLIENT, "scp");
+      log.info("Using {} protocol for file transfer", fileTransferClient);
+      if ("scp".equals(fileTransferClient)) {
+         this.client = new Scp(session);
+      } else if ("sftp".equals(fileTransferClient)) {
+         this.client = new Sftp(session);
+      } else {
+         throw new IllegalArgumentException("Invalid file transfer client. You provided: " + fileTransferClient);
+      }
+   }
+
+   public void upload(String local, String remote) throws IOException {
+      this.upload(List.of(local), remote);
+   }
+
+   public void upload(List<String> local, String remote) throws IOException {
+      this.client.upload(local, remote);
+   }
+
+   @Override
+   public void close() throws IOException {
+      this.client.close();
+   }
+
+   private interface IFileTransfer {
+
+      void upload(List<String> local, String remote) throws IOException;
+
+      void close() throws IOException;
+   }
+
+   private static class Scp implements IFileTransfer {
+
+      private ScpClient scpClient;
+
+      Scp(ClientSession session) {
+         this.scpClient = ScpClientCreator.instance().createScpClient(session);
+      }
+
+      @Override
+      public void upload(List<String> local, String remote) throws IOException {
+         if (!local.isEmpty()) {
+            this.scpClient.upload(local.toArray(String[]::new), remote, ScpClient.Option.PreserveAttributes);
+         }
+      }
+
+      @Override
+      public void close() {
+
+      }
+   }
+
+   private static class Sftp implements IFileTransfer {
+
+      private SftpClient sftpClient;
+
+      Sftp(ClientSession session) throws Exception {
+         try {
+            this.sftpClient = SftpClientFactory.instance().createSftpClient(session);
+         } catch (IOException e) {
+            throw new Exception("Problem while creating the Sftp instance", e);
+         }
+      }
+
+      @Override
+      public void upload(List<String> local, String remote) throws IOException {
+         for (String l : local) {
+            try (InputStream in = new FileInputStream(l);
+                  OutputStream out = sftpClient.write(remote + "/" + new File(l).getName())) {
+               byte[] buffer = new byte[8192];
+               int len;
+               while ((len = in.read(buffer)) != -1) {
+                  out.write(buffer, 0, len);
+               }
+            }
+         }
+      }
+
+      @Override
+      public void close() throws IOException {
+         this.sftpClient.close();
+      }
+   }
+}
--- a/clustering/src/main/java/io/hyperfoil/deploy/ssh/FileTransfer.java
+++ b/clustering/src/main/java/io/hyperfoil/deploy/ssh/FileTransfer.java
@@ -93,7 +93,7 @@ public class FileTransfer implements AutoCloseable {
          for (String l : local) {
             try (InputStream in = new FileInputStream(l);
                   OutputStream out = sftpClient.write(remote + "/" + new File(l).getName())) {
-               byte[] buffer = new byte[8192];
+               byte[] buffer = new byte[32768];
                int len;
                while ((len = in.read(buffer)) != -1) {
                   out.write(buffer, 0, len);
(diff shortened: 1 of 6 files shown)


clustering/src/main/java/io/hyperfoil/deploy/ssh/FileTransfer.java
```diff
@@ -0,0 +1,110 @@
...
+            }
+         }
+      }
+
+      @Override
+      public void close() throws IOException {
+         this.sftpClient.close();
+      }
+   }
+}
```
stalep, line 128: Suggestion: refactor `FileTransfer` to be an abstract class with a static factory method, eliminating the `IFileTransfer` interface and the delegation layer.

Benefits:
- Removes `IFileTransfer` (and its non-idiomatic `I` prefix)
- `Scp` and `Sftp` extend `FileTransfer` directly - no wrapping
- Factory method returns `IOException` instead of broad `Exception`
- Fields are `final`
- SFTP buffer increased to 32KB for better throughput on larger JARs

```java
public abstract class FileTransfer implements AutoCloseable {

   private static final Logger log = LogManager.getLogger(FileTransfer.class);

   public static FileTransfer create(ClientSession session) throws IOException {
      String protocol = Properties.get(Properties.FILE_TRANSFER_CLIENT, "scp");
      log.info("Using {} protocol for file transfer", protocol);
      return switch (protocol) {
         case "scp" -> new Scp(session);
         case "sftp" -> new Sftp(session);
         default -> throw new IllegalArgumentException(
               "Invalid file transfer client. You provided: " + protocol + ". Supported: scp, sftp");
      };
   }

   public void upload(String local, String remote) throws IOException {
      upload(List.of(local), remote);
   }

   public abstract void upload(List<String> local, String remote) throws IOException;

   @Override
   public abstract void close() throws IOException;

   static class Scp extends FileTransfer {

      private final ScpClient scpClient;

      Scp(ClientSession session) {
         this.scpClient = ScpClientCreator.instance().createScpClient(session);
      }

      @Override
      public void upload(List<String> local, String remote) throws IOException {
         if (!local.isEmpty()) {
            this.scpClient.upload(local.toArray(String[]::new), remote, ScpClient.Option.PreserveAttributes);
         }
      }

      @Override
      public void close() {
         // ScpClient has no close method
      }
   }

   static class Sftp extends FileTransfer {

      private final SftpClient sftpClient;

      Sftp(ClientSession session) throws IOException {
         this.sftpClient = SftpClientFactory.instance().createSftpClient(session);
      }

      @Override
      public void upload(List<String> local, String remote) throws IOException {
         for (String l : local) {
            String target = remote + "/" + new File(l).getName();
            try (InputStream in = new FileInputStream(l);
                  OutputStream out = sftpClient.write(target)) {
               byte[] buffer = new byte[32768];
               int len;
               while ((len = in.read(buffer)) != -1) {
                  out.write(buffer, 0, len);
               }
            }
         }
      }

      @Override
      public void close() throws IOException {
         this.sftpClient.close();
      }
   }
}
```

clustering/src/main/java/io/hyperfoil/deploy/ssh/SshDeployedAgent.java
```diff
@@ -97,9 +100,12 @@ public void stop() {
...
       this.session = session;
       this.exceptionHandler = exceptionHandler;
-
-      this.scpClient = ScpClientCreator.instance().createScpClient(session);
-
+      try {
+         this.fileTransfer = new FileTransfer(session);
+      } catch (Exception e) {
+         exceptionHandler.accept(new DeploymentException("Failed to initialize file transfer", e));
+         return;
```
stalep, line 107: With the factory method approach, this becomes:

```java
      try {
         this.fileTransfer = FileTransfer.create(session);
      } catch (IOException e) {
         exceptionHandler.accept(new DeploymentException("Failed to initialize file transfer", e));
         return;
      }
```

Narrowing the catch to `IOException` instead of `Exception`.

clustering/src/main/java/io/hyperfoil/deploy/ssh/SshDeployedAgent.java
```diff
@@ -205,7 +215,7 @@ public void close() throws IOException {
          String filename = log4jConfigurationFile.substring(log4jConfigurationFile.lastIndexOf(File.separatorChar) + 1);
```
stalep, line 215: **Bug:** This passes `targetFile` which is already a full path including the filename (e.g. `/dir/agentlib/log4j2.xml`). But `FileTransfer.upload(String, String)` delegates to `upload(List.of(local), remote)`, and the SFTP implementation does `remote + "/" + new File(l).getName()` - resulting in `/dir/agentlib/log4j2.xml/log4j2.xml`.

The SCP path works because the underlying `ScpClient.upload(String[], String, Option...)` handles single-file-to-file-path correctly.

You need either:
1. A separate `upload` method for single-file-to-exact-path (no filename appending), or
2. Have the SFTP implementation detect whether `remote` is a directory or a file path

Option 1 is simpler - add a boolean parameter or a separate method name like `uploadTo` for the exact-path case.

test-suite/src/test/java/io/hyperfoil/deploy/ssh/FileTransferTest.java
```diff
@@ -0,0 +1,187 @@
...
+ * <p>
+ * Set these system properties to configure the test:
+ * - ssh.test.host (default: localhost)
+ * - ssh.test.port (default: 22)
+ * - ssh.test.user (default: current user)
+ * - io.hyperfoil.file.transfer.client (default: scp) - set to "scp" or "sftp"
+ * <p>
+ * Example usage:
+ * mvn test -Dtest=FileTransferPerformanceTest -Dio.hyperfoil.file.transfer.client=scp
+ * mvn test -Dtest=FileTransferPerformanceTest -Dio.hyperfoil.file.transfer.client=sftp
```
stalep: Javadoc references the wrong class name.

Should be `FileTransferTest` instead of `FileTransferPerformanceTest`.

test-suite/src/test/java/io/hyperfoil/deploy/ssh/FileTransferTest.java
```diff
@@ -0,0 +1,187 @@
...
+      log.info("Protocol: {}", protocol);
+      log.info("Host: {}@{}:{}", TEST_USER, TEST_HOST, TEST_PORT);
+      log.info("Remote directory: {}", remoteDir);
+      log.info("Seed: " + SEED);
+
+      int[] fileCounts = { 1, 10, 100, 1000 };
+
+      for (int fileCount : fileCounts) {
+         log.info("--- Testing with {} file(s) ---", fileCount);
+
```
stalep: **Bug:** The `remoteDir` is not cleaned between iterations. After the 1-file iteration uploads 1 file, the 10-file iteration will find 11 files in `remoteDir`, causing `assertEquals(files.size(), uploadedFileCount)` to fail.

Add cleanup at the start of each iteration:

```java
         // Clean remote directory from previous iteration
         try (var files = Files.list(remoteDir)) {
            files.forEach(p -> {
               try {
                  Files.deleteIfExists(p);
               } catch (IOException e) {
                  log.warn("Failed to clean up {}", p, e);
               }
            });
         }
```

test-suite/src/test/java/io/hyperfoil/deploy/ssh/FileTransferTest.java
```diff
@@ -0,0 +1,187 @@
...
+         }
+
+         files.add(file.getAbsolutePath());
+      }
+
+      return files;
+   }
+
+   private long testFileTransfer(List<String> files) throws Exception {
+      try (FileTransfer fileTransfer = new FileTransfer(session)) {
```
stalep, line 31: With the refactor this becomes `FileTransfer.create(session)`.

stalep (changes requested):
Nice work on the performance improvement - the SCP vs SFTP numbers are compelling, especially for multi-file transfers.

A few suggestions on the design and two bugs I spotted. The main design suggestion is to make `FileTransfer` an abstract class instead of using delegation through `IFileTransfer`. This eliminates the interface (whose `I` prefix is not idiomatic Java), removes the delegation boilerplate, and lets the factory logic live in a static method with a proper return type.

I also noticed a bug in the SFTP single-file upload path and a test bug with missing cleanup between iterations - details in the inline comments.

franz1981:
@diegolovison this is still valid or https://github.com/Hyperfoil/Hyperfoil/pull/780 is enough and we can close it?
Asking to save others (thanks @stalep to have reviewed it the same!) to review it accidentaly

diegolovison:
@franz1981 Yes. This PR is valid.

diegolovison:
@stalep I decided to keep the inner interface and inner implementation to reduce the surface of the change.
It is ready to review again

stalep:
I think this looks good and can be merged as it. I personally do not like the IFileTransfer, but I guess that's a more personal take.

Issue #790, opened 2026-05-15
wrk2 0.29.1 calibration failure


franz1981:
## Summary

`jbang wrk2@hyperfoil` (resolves to Hyperfoil 0.29.1) fails calibration with any combination of rate and connections > 1. Hyperfoil 0.28.0 works fine with the same setup.

## Reproducer

```bash
# Download a Loom-enabled JDK (required by the test server)
# Get the latest build from https://builds.shipilev.net/openjdk-jdk-loom/
# Or use the Docker image: shipilev/openjdk:loom
export JAVA_HOME=/path/to/loom-jdk

# Clone and build the test server
git clone https://github.com/franz1981/Netty-VirtualThread-Scheduler.git
cd Netty-VirtualThread-Scheduler
mvn -B package -DskipTests

# Start the server (uses epoll, mockless mode — no external dependencies)
java --enable-preview \
  -Djdk.virtualThreadScheduler.implClass=io.netty.loom.spi.NettyScheduler \
  -cp benchmark-runner/target/benchmark-runner.jar \
  io.netty.loom.benchmark.runner.HandoffHttpServer \
  --port 8081 --threads 2 --io epoll --mode NETTY_SCHEDULER --mockless

# Verify the server works
curl http://localhost:8081/health   # should return "OK"
curl http://localhost:8081/fruits   # should return JSON

# FAILS: wrk2 0.29.1 (default via jbang catalog)
jbang wrk2@hyperfoil -t1 -c2 -d10s -R500 http://localhost:8081/fruits
# Output: Error: Missing Statistics for 'test'. Found only for: calibration

# Also fails with any combination:
jbang wrk2@hyperfoil -t4 -c100 -d30s -R60000 http://localhost:8081/fruits
jbang wrk2@hyperfoil -t2 -c10 -d30s -R10000 http://localhost:8081/fruits
jbang wrk2@hyperfoil -t1 -c10 -d30s -R5000 http://localhost:8081/fruits

# WORKS: 1 connection only
jbang wrk2@hyperfoil -t1 -c1 -d10s -R100 http://localhost:8081/fruits

# WORKS: wrk (not wrk2, max throughput mode)
jbang wrk@hyperfoil -t2 -c100 -d10s http://localhost:8081/fruits

# WORKS: wrk2 0.28.0
jbang run --main io.hyperfoil.cli.commands.Wrk2 \
  --deps io.hyperfoil:hyperfoil-cli:0.28.0 \
[...]

PR #791, opened 2026-05-15
Prevent log spam during session reset

diegolovison:
Prevent log spam during session reset

Found during the troubleshooting of https://github.com/Hyperfoil/Hyperfoil/issues/790

### DIFF
--- a/api/src/main/java/io/hyperfoil/api/session/PhaseInstance.java
+++ b/api/src/main/java/io/hyperfoil/api/session/PhaseInstance.java
@@ -70,4 +70,6 @@ public interface PhaseInstance {
          return this.ordinal() >= RUNNING.ordinal();
       }
    }
+
+   void notifyRequestsInFlight();
 }
--- a/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
@@ -44,6 +44,7 @@ public abstract class PhaseInstanceImpl implements PhaseInstance {
    private volatile Throwable error;
    private volatile boolean sessionLimitExceeded;
    private Runnable failedSessionAcquisitionAction;
+   private final AtomicInteger requestsInFlight = new AtomicInteger(0);
 
    public static PhaseInstance newInstance(Phase def, String runId, int agentId) {
       PhaseCtor ctor = constructors.get(def.model.getClass());
@@ -296,6 +297,13 @@ public abstract class PhaseInstanceImpl implements PhaseInstance {
          sessionLimitExceeded = true;
       }
       notifyFinished(null);
+   }
+
+   @Override
+   public void notifyRequestsInFlight() {
+      if (this.requestsInFlight.getAndIncrement() == 0) {
+         log.warn("#Phase {} completed with requests in-flight!", this.getName());
+      }
    }
 
    public static class AtOnce extends PhaseInstanceImpl {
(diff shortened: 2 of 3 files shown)


franz1981 (changes requested):
I would instead replace it with a summary - we don't need at all to know each single session ID, unless we are debugging.
So; debug enabled, logging singularly each session. if not enabled, a single warn with the comulative statistics (no per session).

diegolovison:
> a single warn

It requires storing the run/phase in a concurrent structure. I don't think we need this kind of control.

Also, this can be on the hot path. If knowing in-flight is good, we can track it as a metric. ( I don't think we need / because this will show less throughput)

diegolovison:
How it was before:
```
From b34f67ed4b943d3979aaabbc1eeefcb904d135c1 Mon Sep 17 00:00:00 2001
From: Diego Lovison <diegolovison@gmail.com>
Date: Fri, 15 May 2026 15:49:42 -0300
Subject: [PATCH] Prevent log spam during session reset

---
 .../io/hyperfoil/http/HttpRequestPool.java    | 25 +++++++++++--------
 1 file changed, 15 insertions(+), 10 deletions(-)

diff --git a/http/src/main/java/io/hyperfoil/http/HttpRequestPool.java b/http/src/main/java/io/hyperfoil/http/HttpRequestPool.java
index 9b7853e44..15cb3fc81 100644
--- a/http/src/main/java/io/hyperfoil/http/HttpRequestPool.java
+++ b/http/src/main/java/io/hyperfoil/http/HttpRequestPool.java
@@ -25,16 +25,21 @@ public static LimitedPool<HttpRequest> get(Session session) {
    @Override
    public void onSessionReset(Session session) {
       if (!isFull()) {
-         // We can't guarantee that requests will be back in session's requestPool when it terminates
-         // because if the requests did timeout (calling handlers and eventually letting the session terminate)
-         // it might still be held in the connection.
-         for (HttpRequest request : (HttpRequest[]) originalObjects) {
-            // We won't issue the warning for invalid requests because these are likely not in flight anymore
-            // and we are stopping the session exactly due to the invalid request.
-            if (!request.isCompleted() && request.isValid()) {
-               log.warn("#Phase {}/{} with session {} completed with requests in-flight!",
-                     session.phase().getName(), session.phase().status(), session.uniqueId());
-               break;
+         // This is a hot path when the scenario ends-up with in-flight requests.
+         // Guarded with isDebugEnabled to prevent log spam from excessive warnings during failures.
+         // Ideally, dropped in-flight requests should be tracked as metrics/statistics rather than logged.
+         if (log.isDebugEnabled()) {
+            // We can't guarantee that requests will be back in session's requestPool when it terminates
+            // because if the requests did timeout (calling handlers and eventually letting the session terminate)
+            // it might still be held in the connection.
+            for (HttpRequest request : (HttpRequest[]) originalObjects) {
+               // We won't issue the warning for invalid requests because these are likely not in flight anymore
+               // and we are stopping the session exactly due to the invalid request.
+               if (!request.isCompleted() && request.isValid()) {
+                  log.warn("#Phase {}/{} with session {} completed with requests in-flight!",
+                        session.phase().getName(), session.phase().status(), session.uniqueId());
+                  break;
+               }
             }
          }
          cancelRequests();
```

diegolovison:
@franz1981, how about now?
Instead of a gazillion entries in the logs, now we have:
```
18:31:53,611 INFO  (main) [i.h.c.i.SimulationRunner] Simulation initialization took 61 ms
18:32:19,750 WARN  (epollEventLoopGroup-2-1) [i.h.c.i.PhaseInstanceImpl] #Phase test completed with requests in-flight!
18:32:19,755 INFO  (main) [i.h.s.WrkScenarioTest] Test duration: 26s
```

franz1981 (approved):
well done!

PR #792, opened 2026-05-16
Bump Vert.x to 4.5.26

diegolovison:
Bump Vert.x to 4.5.26

### DIFF
--- a/clustering/src/main/java/io/hyperfoil/clustering/AgentVerticle.java
+++ b/clustering/src/main/java/io/hyperfoil/clustering/AgentVerticle.java
@@ -49,7 +49,7 @@ public class AgentVerticle extends AbstractVerticle {
 
    @Override
    public void start() {
-      deploymentId = deploymentID();
+      String vertxDeploymentId = deploymentID();
       name = context.config().getString("name");
       if (name == null) {
          name = Properties.get(Properties.AGENT_NAME, null);
@@ -59,7 +59,7 @@ public class AgentVerticle extends AbstractVerticle {
             name = InetAddress.getLocalHost().getHostName();
          } catch (UnknownHostException e) {
             log.debug("Cannot deduce name from host name", e);
-            name = deploymentId;
+            name = vertxDeploymentId;
          }
       }
       runId = context.config().getString("runId");
@@ -71,10 +71,21 @@ public class AgentVerticle extends AbstractVerticle {
       }
       eb = vertx.eventBus();
 
+      if (vertx.isClustered()) {
+         if (vertx instanceof VertxInternal) {
+            nodeId = ((VertxInternal) vertx).getClusterManager().getNodeId();
+         }
+      }
+      // Vert.x 4.5.26+ deployment IDs are local to a JVM in this setup; agent control addresses must be cluster-wide unique.
+      deploymentId = agentControlAddress(vertxDeploymentId);
+      log.info("Agent {} for run {} uses control address {} (Vert.x deployment ID {}, node ID {})", name, runId,
+            deploymentId, vertxDeploymentId, nodeId);
+
       eb.consumer(deploymentId, message -> {
          try {
             AgentControlMessage controlMessage = (AgentControlMessage) message.body();
             if (controlMessage == null) {
+               log.error("{} Failed to decode message body", deploymentId);
                message.fail(1, "Could not decode message body. Does this Agent have the same version as the Controller?");
                return;
             }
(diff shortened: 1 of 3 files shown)


diegolovison:
JDK26 is expected to fail
I wish to keep the info messages. They helped a lot.

franz1981 (changes requested):
Two issues:

1. `eb.request(Feeds.RESPONSE, new PhaseChangeMessage(...))` (L277) will always time out — the RESPONSE consumer in ControllerVerticle never replies. Master used `eb.send` here. Either revert to `eb.send` or add `message.reply(...)` in the controller's RESPONSE handler.

2. `log.info("Sending AgentHello...")` (L99) fires every second in `setPeriodic(1000, ...)` — should be `log.debug`.

diegolovison:
1) I didn't change this. If it is an issue, please raise one

> eb.request(Feeds.RESPONSE, new PhaseChangeMessage(...)) (L277) will always time out — the RESPONSE consumer in ControllerVerticle never replies. Master used eb.send here. Either revert to eb.send or add message.reply(...) in the controller's RESPONSE handler.


2) Done

clustering/src/main/java/io/hyperfoil/clustering/AgentVerticle.java
```diff
@@ -289,9 +307,9 @@ private void initBenchmark(Benchmark benchmark, int agentId) {
 
       runner.openConnections(vertx::executeBlocking, result -> {
          if (result.succeeded()) {
-            eb.send(Feeds.RESPONSE, new AgentReadyMessage(deploymentID(), runId));
+            eb.send(Feeds.RESPONSE, new AgentReadyMessage(deploymentId, runId));
```
franz1981, line 310: what's going on with the original `deploymentID()` method? It is still there, I assume - so what's the difference between this local value and using that one?

franz1981 (changes requested):

clustering/src/main/java/io/hyperfoil/clustering/AgentVerticle.java
diegolovison: https://github.com/Hyperfoil/Hyperfoil/pull/792/changes#diff-138b2a7ea60157c420ef4e96b492a71468ea329e23cd525ef8063c964316e08bR79

I think Vert.x fixed something that introduces this regression.

If you have another idea about how to get the unique `deploymentId` please let me know.

franz1981: which regression? you have a failing test?

diegolovison: Yes. The test `ClusterTestCase.java` started failing after the Vert.x minor upgrade.
* Checkout master branch
* Change to `<version.vertx>4.5.26</version.vertx>`
* Run `rm -fR /tmp/hyperfoil/ && mvn clean package -Pbenchmark -pl test-suite -am -Dtest=ClusterTestCase -Dsurefire.failIfNoSpecifiedTests=false`

franz1981: > If you have another idea about how to get the unique deploymentId please let me know.

Let's ping @wburns or @jabolina so we can ask them to lend an hand on this?

jabolina: Seems to make sense. There was this change: https://github.com/eclipse-vertx/vert.x/commit/70b889f50c65eac4662ad18aef3fa163e082d868

Maybe that should be an `||` instead of `&&`? I am not completely familiar. We can either:

1. Create the unique ID manually as Diego did;
2. Update in `Hyperfoil#clusteredVertx` method to do something like (around lines 54):
```java
      VertxOptions options = new VertxOptions();
      options.setHAEnabled(true);
```

With 2., Vert.x is creating the UUID for us and it wouldn't be duplicated. But note that I am not really familiar how setting HA=true could affect Hyperfoil :(

diegolovison: If you don't see any problem with the current implementation, I think we can stick with it. This prevents other side effects that can occur because of `HA=true`

jabolina: Yeah, this should be good :+1:
The generated ID seems unique with the agent name and nodeId.

diegolovison: @franz1981 this is ready

franz1981:
@diegolovison can you repush to see the JDK 22 run too?

diegolovison:
@franz1981 done

PR #797, opened 2026-05-18
Check if trace is enabled before log.trace

diegolovison:
Check if trace is enabled before log.trace


franz1981 (commented):
This is not needed, the log API already handle this right: why you want to change it?

diegolovison:
Forgot about it. Thanks.

PR #798, opened 2026-05-18
Prevent log spam if request is not valid

diegolovison:
Prevent log spam if request is not valid


franz1981 (changes requested):
The INFO→DEBUG change for StopOnInvalidAction is correct — this message is redundant with the per-request WARN from RangeStatusValidator.handleStatus() (RangeStatusValidator.java:30). 
But **that** WARN is the bigger spam source and isn't addressed here. 
Same scenario produces matching WARN volume that remains after this PR. 
Address RangeStatusValidator WARN here or in a follow-up.

franz1981 (approved):

franz1981 (approved):

PR #799, opened 2026-05-18
Fix NPE during pulse

### LINKED ISSUE #638: NPE: Cannot read field "session" because "this.request" is null
**Describe the bug**
Set `maxDuration` as `durantion` when the rate is super high, and the SUT is super slow

```
java.lang.NullPointerException: Cannot read field "session" because "this.request" is null
	at io.hyperfoil.http.steps.HttpRequestContext.accept(HttpRequestContext.java:37)
	at io.hyperfoil.http.connection.SharedConnectionPool.pulse(SharedConnectionPool.java:201)
	at io.hyperfoil.http.connection.SharedConnectionPool.handleNewConnection(SharedConnectionPool.java:323)
	at io.hyperfoil.http.connection.Http1xConnection.checkActivated(Http1xConnection.java:75)
	at io.hyperfoil.http.connection.Http1xConnection.channelActive(Http1xConnection.java:69)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelActive(AbstractChannelHandlerContext.java:260)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelActive(AbstractChannelHandlerContext.java:238)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelActive(AbstractChannelHandlerContext.java:231)
```

**To Reproduce**
The `waiting` deque should have items
Image

**Expected behavior**
No NPE

**Additional context**

diegolovison:
Fixes: https://github.com/Hyperfoil/Hyperfoil/issues/638

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
@@ -103,7 +103,7 @@ public abstract class PhaseInstanceImpl implements PhaseInstance {
          status = Status.RUNNING;
       }
       recordAbsoluteStartTime();
-      log.debug("{} changing status to RUNNING", def.name);
+      log.debug("Phase '{}' changing status to RUNNING", def.name);
       phaseChangeHandler.onChange(def, Status.RUNNING, false, error)
             .thenRun(this::proceed);
    }
@@ -125,7 +125,7 @@ public abstract class PhaseInstanceImpl implements PhaseInstance {
       synchronized (this) {
          if (status == Status.RUNNING) {
             status = Status.FINISHED;
-            log.debug("{} changing status to FINISHED", def.name);
+            log.debug("Phase '{}' changing status to FINISHED", def.name);
          } else {
             log.debug("{} already in state {}, not finishing", def.name, status);
          }
@@ -159,7 +159,7 @@ public abstract class PhaseInstanceImpl implements PhaseInstance {
             status = Status.TERMINATING;
          }
       }
-      log.debug("{} changing status to TERMINATING", def.name);
+      log.debug("Phase '{}' changing status to TERMINATING", def.name);
       tryTerminate();
    }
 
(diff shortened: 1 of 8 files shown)


franz1981 (changes requested):
The `else` branch has three bugs:

1. **Connection leak.** `acquireNow(false)` removed a connection from `available`, incremented `inFlight`/`usedConnections`, called `onAcquire()` (increments `aboutToSend`). None of this is reversed for an invalid consumer. No `cancelAcquire()` exists, so `release()` alone can't undo it.

2. **`waiting.clear()` drops valid consumers.** Each had `blockedSessions.incrementUsed()` — never decremented. Also drops consumers from *other* sessions that are still alive. Test passes only because in `testWrk2HighLoad` all sessions reset simultaneously.

3. **`failures > MAX_FAILURES` path (line 209-212) still calls `consumer.accept(null)` without `isValid()` check** — same NPE.

Suggest checking `isValid()` *before* `acquireNow()` to avoid the acquire-then-release problem entirely:

```java
// Drain invalid consumers first
while ((consumer = waiting.peek()) != null && !consumer.isValid()) {
    waiting.poll();
    blockedSessions.decrementUsed();
}
consumer = waiting.poll();
if (consumer != null) {
    // existing logic unchanged — consumer is guaranteed valid
}
```

This fixes all three issues: no connection leak, correct `blockedSessions` accounting, and the `failures > MAX_FAILURES` path can't reach an invalid consumer.

diegolovison:
@franz1981 This PR is still a draft, and we can talk about the review above later.

I would like to propose this https://github.com/Hyperfoil/Hyperfoil/pull/799/changes/e5b6387628f25b9227c1cf079f96c016e5d29d82#diff-05dbfe9fd8851d72c066f0c8d6944dc5d480a7e96c63b5a251dba53d1efa571dR203-R209

We cannot have NPE only in the logs, and we cannot account for the session (the object that is able to collect statistics).

What's your opinion? With that, we have a real reproducer for the issue.

franz1981:
@diegolovison 

the try-catch + shutdown is a safety net, not a fix, yet...
The else branch still has the logic bugs explained at https://github.com/Hyperfoil/Hyperfoil/pull/799#pullrequestreview-4319732003 but you added a real reproducer test which is super valuable.

diegolovison:
ACK for problem 2), but the suggested code also won't work. It can be valid and "1 time" later be invalid because of the async nature.

diegolovison:
Closed in favor of https://github.com/Hyperfoil/Hyperfoil/pull/816

PR #808, opened 2026-05-19
Add Java 22 to the matrix

diegolovison:
Add Java 22 to the matrix

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


diegolovison:
@franz1981 23, 24, 25, 26 are failing as expected because of the Infinispan version rollback.
Add 22, which is working as a sanity check.

franz1981:
why 22? is not even an LTS?

diegolovison:
Catch errors fast until we have the next LTS working.
If a code change breaks 22, high chance of breaking the next LTS.

franz1981 (approved):

PR #815, opened 2026-05-27
fix: correct lifecycle management for phases without explicit durations

diegolovison:
- Prevent LocalSimulationRunner from prematurely terminating phases with a negative duration.
- Explicitly call finish() in AtOnce and Noop phases to prevent deadlocks.
--- Adding finish() to the proceed() methods of AtOnce and Noop perfectly closes the loophole, allowing them to transition out of the RUNNING state exactly when they are supposed to.
- Refactor phase status transitions into a centralized changeStatus method.
- Add debug logging for session resets.

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
@@ -100,10 +100,9 @@ public abstract class PhaseInstanceImpl implements PhaseInstance {
       this.executorGroup = executorGroup;
       synchronized (this) {
          assert status == Status.NOT_STARTED : "Status is " + status;
-         status = Status.RUNNING;
+         this.changeStatus(Status.RUNNING);
       }
       recordAbsoluteStartTime();
-      log.debug("{} changing status to RUNNING", def.name);
       phaseChangeHandler.onChange(def, Status.RUNNING, false, error)
             .thenRun(this::proceed);
    }
@@ -124,8 +123,7 @@ public abstract class PhaseInstanceImpl implements PhaseInstance {
    public void finish() {
       synchronized (this) {
          if (status == Status.RUNNING) {
-            status = Status.FINISHED;
-            log.debug("{} changing status to FINISHED", def.name);
+            this.changeStatus(Status.FINISHED);
          } else {
             log.debug("{} already in state {}, not finishing", def.name, status);
          }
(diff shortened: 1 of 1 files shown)


diegolovison:
How I found this

```
Index: http/src/main/java/io/hyperfoil/http/steps/HttpRequestContext.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/http/src/main/java/io/hyperfoil/http/steps/HttpRequestContext.java b/http/src/main/java/io/hyperfoil/http/steps/HttpRequestContext.java
--- a/http/src/main/java/io/hyperfoil/http/steps/HttpRequestContext.java	(revision a0872d85a552d9e84c8042fc83d7179752c4c865)
+++ b/http/src/main/java/io/hyperfoil/http/steps/HttpRequestContext.java	(date 1779895006582)
@@ -1,5 +1,8 @@
 package io.hyperfoil.http.steps;

+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
 import io.hyperfoil.api.session.Session;
 import io.hyperfoil.http.api.ConnectionConsumer;
 import io.hyperfoil.http.api.HttpConnection;
@@ -11,6 +14,9 @@
  * state of execution here.
  */
 class HttpRequestContext implements Session.Resource, ConnectionConsumer {
+
+   private static final Logger log = LogManager.getLogger(HttpRequestContext.class);
+
    HttpRequest request;
    HttpConnection connection;
    boolean ready;
@@ -18,6 +24,7 @@

    @Override
    public void onSessionReset(Session session) {
+      log.info("Session reset status: {}", session.phase().status());
       reset();
    }
```

Then `mvn clean package -pl http -am -Dtest=ChunkedTransferTest#testRandomCutBuffers -Dsurefire.failIfNoSpecifiedTests=false`

Look at `/home/dlovison/github/Hyperfoil/Hyperfoil/http/target/surefire-reports/io.hyperfoil.http.ChunkedTransferTest-output.txt`


You will see that the output is matching `assertThat(counter.get()).isEqualTo(16 * 64);`
BUT
the status must be something like `RUNNING`

franz1981 (changes requested):
This change per se is good, but it is making `ChunkedTransferTest#testRandomCutBuffers` to have a huge increase in [exec loop iterations](https://github.com/diegolovison/Hyperfoil/blob/5965ee0dbc08269619e0cc64dde30852aff30bbc/core/src/main/java/io/hyperfoil/core/impl/LocalSimulationRunner.java#L83) because nextPhaseFinish computes startTime + (-1) for the Sequentially phase that now stays RUNNING, producing a negative delay every iteration

franz1981 (approved):
LGTM!

Issue #643, opened 2025-12-29
Connection Leak During PhaseInstance.Status.TERMINATING


diegolovison:
**Describe the bug**
There are issues with connection management when a phase enters PhaseInstance.Status.TERMINATING state. The problems begin from the split responsibility between connection acquisition and release, combined with the `stop` mechanism during termination.

* PrepareHttpRequestStep - Acquires connection via `connectionPool.acquire(false, context)`
* SendHttpRequestStep - Sends request via `request.send(context.connection, ...)`

The connection is released later when the response completes via `releasePoolAndPulse()`

**Problem 1: Connection Acquired But Request Not Sent**

This can trigger TERMINATING check before SendHttpRequestStep runs. Connection is stored in context but never used or released.

``` 
PrepareHttpRequestStep executes
    ↓
connectionPool.acquire() called [line 89]
    ↓
Connection acquired and stored in context
    ↓
context.accept(connection) calls session.proceed() [line 40]
    ↓
Session checks TERMINATING status [line 265]
    ↓
Session calls stop() [line 269]
    ↓
SendHttpRequestStep NEVER executes
    ↓
Connection NEVER released - LEAK!
```

**Problem 2: Request sent but response handling interrupted**

```
SendHttpRequestStep executes
    ↓
Request sent to server
    ↓
Response starts arriving
    ↓
Phase enters TERMINATING
    ↓
Session.proceed() checks TERMINATING [line 265]
    ↓
Session stops [line 269]
    ↓
Response handler may or may not complete
    ↓
releasePoolAndPulse() may or may not be called
    ↓
Connection may be leaked OR released
```

**To Reproduce**
https://github.com/Hyperfoil/Hyperfoil/pull/641

diegolovison:
Fixed https://github.com/Hyperfoil/Hyperfoil/pull/816

Issue #642, opened 2025-12-29
Pulse Continues During PhaseInstance.Status.TERMINATING


diegolovison:
**Describe the bug**
When a phase enters PhaseInstance.Status.TERMINATING, there's a race condition and an infinite loop between:
* `PhaseInstanceImpl.tryTerminate()` - Forces blocked sessions to proceed by calling `session.proceed()`
* `SessionImpl.proceed()` - Checks if phase is TERMINATING and calls stop()
* `SharedConnectionPool.pulse()` - Gives connections to waiting sessions and schedules more pulses

---

* `Sessions` are waiting for connections in the `waiting` queue
* `tryTerminate()` calls `session.proceed()` to unblock them
* But `pulse()` also gives them connections via `consumer.accept(connection)`. If the connection is null, it's adding `consumer` back to the waiting queue.

**To Reproduce**
https://github.com/Hyperfoil/Hyperfoil/pull/641

franz1981:
What is the high level effect to user? Do you have a test which shows what this behaviour cause?

diegolovison:
Fixed https://github.com/Hyperfoil/Hyperfoil/pull/816

PR #816, opened 2026-05-29
fix(http): handle resource cleanup and prevent NPE during phase termination

diegolovison:
- Fixes a NullPointerException in HttpRequestContext where `this.request` is null when acquiring a connection.
- Stops the connection pool pulse loop when a phase is terminating by introducing `Session.tryTerminate()`.
- Prevents connection leaks during termination by clearing the waiting queue and releasing active connections back to the pool.

Fixes: 
- https://github.com/Hyperfoil/Hyperfoil/issues/643
- https://github.com/Hyperfoil/Hyperfoil/issues/642
- https://github.com/Hyperfoil/Hyperfoil/issues/638

Stacktrace before the fix
```
09:21:46,049 WARN  (epollEventLoopGroup-2-1) [i.h.h.c.Http1xConnection] Exception in Http1xConnection{/127.0.0.1:46166 -> localhost/127.0.0.1:41281, status=OPEN, size=0+1:[]}java.lang.NullPointerException: Cannot read field "session" because "this.request" is null
	at io.hyperfoil.http.steps.HttpRequestContext.accept(HttpRequestContext.java:37)
	at io.hyperfoil.http.connection.SharedConnectionPool.pulse(SharedConnectionPool.java:203)
	at io.hyperfoil.http.connection.SharedConnectionPool.handleNewConnection(SharedConnectionPool.java:325)
	at io.hyperfoil.http.connection.Http1xConnection.checkActivated(Http1xConnection.java:75)
	at io.hyperfoil.http.connection.Http1xConnection.channelActive(Http1xConnection.java:69)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelActive(AbstractChannelHandlerContext.java:260)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelActive(AbstractChannelHandlerContext.java:238)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelActive(AbstractChannelHandlerContext.java:231)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelActive(DefaultChannelPipeline.java:1345)
```

### DIFF
--- a/http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
+++ b/http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
@@ -52,6 +52,7 @@ class SharedConnectionPool extends ConnectionPoolStats implements HttpConnection
    private final Deque<ConnectionConsumer> waiting = new ArrayDeque<>();
    private ScheduledFuture<?> pulseFuture;
    private ScheduledFuture<?> keepAliveFuture;
+   private boolean shouldPulse = true;
 
    SharedConnectionPool(HttpClientPoolImpl clientPool, EventLoop eventLoop, ConnectionPoolConfig sizeConfig) {
       super(clientPool.authority);
@@ -139,13 +140,14 @@ class SharedConnectionPool extends ConnectionPoolStats implements HttpConnection
          log.trace("Release {} (became available={} after request={})", connection, becameAvailable, afterRequest);
       }
       if (becameAvailable) {
-         assert !connection.isClosed();
-         if (connection.inFlight() == 0) {
-            // We are adding to the beginning of the queue to prefer reusing connections rather than cycling
-            // too many often-idle connections
-            available.addFirst(connection);
-         } else {
-            available.addLast(connection);
+         if (!connection.isClosed()) {
+            if (connection.inFlight() == 0) {
+               // We are adding to the beginning of the queue to prefer reusing connections rather than cycling
+               // too many often-idle connections
+               available.addFirst(connection);
+            } else {
+               available.addLast(connection);
+            }
          }
       }
       if (afterRequest) {
@@ -192,8 +194,23 @@ class SharedConnectionPool extends ConnectionPoolStats implements HttpConnection
 
    @Override
    public void pulse() {
+      assert executor().inEventLoop();
       if (trace) {
          log.trace("Pulse to {} ({} waiting)", authority, waiting.size());
+      }
+      // session terminated, there is nothing that we can do
+      if (!shouldPulse) {
+         blockedSessions.decrementUsed(waiting.size());
+         assert blockedSessions.current() == 0;
+         waiting.clear();
+         // make the connections available again
+         for (HttpConnection conn : connections) {
+            if (!available.contains(conn)) {
+               release(conn, true, false);
+            }
+         }
+         shouldPulse = true;
+         return;
       }
       ConnectionConsumer consumer = waiting.poll();
       if (consumer != null) {
@@ -357,5 +374,10 @@ class SharedConnectionPool extends ConnectionPoolStats implements HttpConnection
             conn.context().flush();
          }
       });
+   }
+
+   @Override
+   public void onSessionTryTerminate() {
+      this.shouldPulse = false;
    }
 }
--- a/http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
+++ b/http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
@@ -378,6 +378,9 @@ class SharedConnectionPool extends ConnectionPoolStats implements HttpConnection
 
    @Override
    public void onSessionTryTerminate() {
+      if (!executor().inEventLoop()) {
+         executor().execute(this::onSessionTryTerminate);
+      }
       this.shouldPulse = false;
    }
 }
--- a/http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
+++ b/http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
@@ -136,6 +136,7 @@ class SharedConnectionPool extends ConnectionPoolStats implements HttpConnection
 
    @Override
    public void release(HttpConnection connection, boolean becameAvailable, boolean afterRequest) {
+      assert executor().inEventLoop();
       if (trace) {
          log.trace("Release {} (became available={} after request={})", connection, becameAvailable, afterRequest);
       }
--- a/http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
+++ b/http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
@@ -396,6 +396,7 @@ class SharedConnectionPool extends ConnectionPoolStats implements HttpConnection
    public void onSessionTryTerminate() {
       if (!executor().inEventLoop()) {
          executor().execute(this::onSessionTryTerminate);
+         return;
       }
       this.shouldPulse = false;
    }
--- a/http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
+++ b/http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
@@ -88,7 +88,10 @@ public class SharedConnectionPool extends ConnectionPoolStats implements HttpCon
                   usedConnections.incrementUsed();
                }
                connection.onAcquire();
-
+               if (log.isDebugEnabled()) {
+                  log.debug("acquireNow: acquired {} to {}, inFlight={} (after), connection.inFlight={}",
+                        connection, authority, inFlight.current(), connection.inFlight());
+               }
                return connection;
             } else {
                availableClosed--;
@@ -105,6 +108,10 @@ public class SharedConnectionPool extends ConnectionPoolStats implements HttpCon
 
    @Override
    public void acquire(boolean exclusiveConnection, ConnectionConsumer consumer) {
+      if (log.isDebugEnabled()) {
+         log.debug("acquire: {} exclusiveConnection={}, waiting={}, available={}, inFlight={}",
+               authority, exclusiveConnection, waiting.size(), available.size(), inFlight.current());
+      }
       HttpConnection connection = acquireNow(exclusiveConnection);
       if (connection != null) {
          consumer.accept(connection);
@@ -139,6 +146,16 @@ public class SharedConnectionPool extends ConnectionPoolStats implements HttpCon
       assert executor().inEventLoop();
       if (trace) {
          log.trace("Release {} (became available={} after request={})", connection, becameAvailable, afterRequest);
+      }
+      if (log.isDebugEnabled()) {
+         boolean alreadyInAvailable = available.contains(connection);
+         log.debug(
+               "release: {} becameAvailable={}, afterRequest={}, connection.inFlight={}, inFlight={} (before), available={}, alreadyInAvailable={}",
+               connection, becameAvailable, afterRequest, connection.inFlight(), inFlight.current(), available.size(),
+               alreadyInAvailable);
+         if (becameAvailable && !connection.isClosed() && alreadyInAvailable) {
+            log.debug("release: DUPLICATE! {} already in available queue", connection);
+         }
       }
       if (becameAvailable) {
          if (!connection.isClosed()) {
@@ -199,8 +216,15 @@ public class SharedConnectionPool extends ConnectionPoolStats implements HttpCon
       if (trace) {
          log.trace("Pulse to {} ({} waiting)", authority, waiting.size());
       }
+      if (log.isDebugEnabled()) {
+         log.debug("pulse: {} shouldPulse={}, waiting={}, available={}, inFlight={}",
+               authority, shouldPulse, waiting.size(), available.size(), inFlight.current());
+      }
       // session terminated, there is nothing that we can do
       if (!shouldPulse) {
+         if (log.isDebugEnabled()) {
+            log.debug("pulse: session terminated, releasing all non-available connections, waiting={}", waiting.size());
+         }
 
          blockedSessions.decrementUsed(waiting.size());
          waiting.clear();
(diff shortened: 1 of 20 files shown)


test-suite/src/test/java/io/hyperfoil/scenario/WrkScenarioTest.java
```diff
@@ -33,6 +33,16 @@ public class WrkScenarioTest extends BaseWrkBenchmarkTest {
 
    protected final Logger log = LogManager.getLogger(getClass());
 
+   @Test
+   public void wrk2SuperSlowServer() throws URISyntaxException {
```
franz1981, line 37: which output you have while you run this without your patch and with your patch?

diegolovison: If we run without the patch, it will produce NPE and fail.

franz1981: I will try reproduce it myself, but please in the description print the stack trace with the error too to help reviewers

diegolovison: Done

franz1981:
Two critical issues with the termination mechanism:

1. shouldPulse is a plain boolean written from tryTerminate() (controller thread) but read from pulse() on the Netty event loop. No happens-before. Needs volatile or dispatching via executor().execute(...).
2. SharedConnectionPool is shared by all sessions on the same event loop. onSessionTryTerminate() sets pool-wide shouldPulse = false, and the next pulse() clears the entire waiting queue — including requests from sessions that are still active. The termination mechanism needs to be per-session, not per-pool.

Also: HttpRequestContext.accept() leaks the connection when request == null (should release it back to the pool).
The release() assert-to-if change and NPE null guard are correct individually.

diegolovison:
@franz1981 please review again.

http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
```diff
@@ -358,4 +391,12 @@ public void shutdown() {
          }
       });
    }
+
+   @Override
+   public void onSessionTryTerminate() {
+      if (!executor().inEventLoop()) {
+         executor().execute(this::onSessionTryTerminate);
```
franz1981, line 467:
```suggestion
         executor().execute(this::onSessionTryTerminate);
         return; 
```

http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
```diff
@@ -358,4 +391,12 @@ public void shutdown() {
...
+      }
+      this.shouldPulse = false;
```
franz1981, line 474: The last line runs from controller thread too, do you need a `return`? see https://github.com/Hyperfoil/Hyperfoil/pull/816/changes#r3388597046

http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
```diff
@@ -192,9 +195,39 @@ public EventLoop executor() {
 
    @Override
    public void pulse() {
+      assert executor().inEventLoop();
       if (trace) {
          log.trace("Pulse to {} ({} waiting)", authority, waiting.size());
       }
+      // session terminated, there is nothing that we can do
+      if (!shouldPulse) {
```
franz1981, line 239: SharedConnectionPool is shared across all sessions on the same event loop. 
When any session terminates, `shouldPulse = false` on the shared pool, and `pulse()` removes ALL waiting consumers — because `isValid()` returns ready, which is false for every consumer in the waiting queue (ready is only set in `accept()`, which hasn't been called on waiting consumers). 
Requests from still-active sessions are discarded.

http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
```diff
@@ -192,9 +195,39 @@ public EventLoop executor() {
...
+         }
+
+         blockedSessions.decrementUsed(removedCount);
+
+         if (blockedSessions.current() > 0) {
+            this.pulse();
+            return;
+         }
+
+         for (HttpConnection conn : connections) {
```
franz1981, line 255: For connections with in-flight requests: this adds them to available now, and the response completion path adds them again later. 
ArrayDeque allows duplicates — same connection acquired by two consumers.

franz1981:
The accept() null guard with connection release, and the release() assert-to-if change - is fine.

What needs redesign is the termination mechanism: it needs to be per-session, not per-pool. 
One approach: have pulse() check each waiting consumer's session termination status, rather than using a pool-wide flag.

franz1981 (changes requested):

http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
diegolovison: Done

http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
diegolovison: Done

http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
diegolovison: > SharedConnectionPool is shared across all sessions on the same event loop.

Yes

> When any session terminates, shouldPulse = false on the shared pool

Yes, but the method is called only when `PhaseInstance::tryTerminate` is called and after all sessions have proceeded

> and pulse() removes ALL waiting consumers — because isValid() returns ready, which is false for every consumer in the waiting queue (ready is only set in accept(), which hasn't been called on waiting consumers).

Good catch. I removed the `isValid` and removed all from the waiting queue.

> Requests from still-active sessions are discarded.

The only edge case that I found was during `pulse`. Double-check `SessionImpl::run`. If an active session reaches that method with `PhaseInstance.Status.TERMINATING`, then it will stop and won't run.

http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
diegolovison: After `PhaseInstance.Status.TERMINATING`, the connection won't be added back. This PR also addresses https://github.com/Hyperfoil/Hyperfoil/issues/643

diegolovison:
> What needs redesign is the termination mechanism: it needs to be per-session, not per-pool.
> One approach: have pulse() check each waiting consumer's session termination status, rather than using a pool-wide flag.

It was my initial approach. But this idea, from my findings, will require a big refactor.

@franz1981 the comments were addressed. Please review again.

franz1981:
Two questions on the remaining design:

1. You said onSessionTryTerminate() is only called after all sessions have proceeded. Can you point to where this invariant is enforced? If SharedConnectionPool is shared across all sessions on the event loop, a single session's termination could trigger onSessionTryTerminate() while other sessions are still active.
2. You said connections won't be added back after TERMINATING status. Which code path prevents Http1xConnection.handleResponse() → pool.release() from adding a connection back to available during termination?

If both invariants hold, the current approach is correct. I just want to verify them.

diegolovison:
> You said onSessionTryTerminate() is only called after all sessions have proceeded. Can you point to where this invariant is enforced? If SharedConnectionPool is shared across all sessions on the event loop, a single session's termination could trigger onSessionTryTerminate() while other sessions are still active.

After your question at https://github.com/Hyperfoil/Hyperfoil/pull/823#discussion_r3388582797 this PR requires https://github.com/Hyperfoil/Hyperfoil/pull/823 to be merged first. Then, I will need to update this PR again to address 

```java
for (int i = 0; i < sessionList.size(); i++) {
   Session session = sessionList.get(i);
   // add resources back
   session.tryTerminate();
}
```

> You said connections won't be added back after TERMINATING status. Which code path prevents Http1xConnection.handleResponse() → pool.release() from adding a connection back to available during termination?

It is describe at https://github.com/Hyperfoil/Hyperfoil/issues/643

http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
```diff
@@ -192,9 +194,26 @@ public EventLoop executor() {
...
       }
+      // session terminated, there is nothing that we can do
+      if (!shouldPulse) {
+
+         blockedSessions.decrementUsed(waiting.size());
+         waiting.clear();
+
+         assert blockedSessions.current() == 0;
+
+         for (HttpConnection conn : connections) {
```
franz1981, line 255: Another concern on this...
`release(conn, true, false)` adds the connection to available but does NOT decrement inFlight (because `afterRequest=false`). 
But `acquireNow()` incremented `inFlight` when the connection was acquired. 
If the connection was acquired for a consumer that's now in the cleared waiting queue, `inFlight` remains inflated.

Also: for connections with in-flight HTTP requests that complete after this cleanup, the response path calls `release(conn, true, true)`, adding the connection to `available` a second time (`ArrayDeque` allows duplicates). 

Your claim that "connections won't be added back after TERMINATING" needs to be verified: `Http1xConnection.handleResponse()` doesn't check phase status before calling `pool.release()`.

franz1981 (changes requested):

franz1981 (commented):
The test has two issues:

1. It checks `inFlightCount` inside the [completion handler](https://github.com/Hyperfoil/Hyperfoil/pull/816/files#diff-8d9e82e8f5ac0b7aef22aba9c4e53f2de9f62b9d63f5c2b42c36c2e6f0d62f25R78-R101), which fires [before `releasePoolAndPulse()` decrements `inFlight`](https://github.com/Hyperfoil/Hyperfoil/blob/76625d30/http/src/main/java/io/hyperfoil/http/connection/Http1xResponseHandler.java#L544-L565). The `inFlight == 1` it sees is the normal state at that point in the response pipeline — not evidence of a leak.

2. [`session.tryTerminate()` at line 113](https://github.com/Hyperfoil/Hyperfoil/pull/816/files#diff-8d9e82e8f5ac0b7aef22aba9c4e53f2de9f62b9d63f5c2b42c36c2e6f0d62f25R113) never sets `shouldPulse = false` because [`initForTesting` registers an `HttpDestinationTableImpl` with empty maps](https://github.com/Hyperfoil/Hyperfoil/blob/76625d30/http/src/main/java/io/hyperfoil/http/HttpRunData.java#L88) — the `!shouldPulse` code path is never entered. The test doesn't exercise the race condition it describes.

diegolovison:
@franz1981 Please review again

franz1981:
`HttpRequestContext.accept()` with `request == null` leaks pool-level `inFlight` by 1.

`cancelAcquire()` undoes `conn.onAcquire()` (aboutToSend--) but `release(conn, true, false)` passes `afterRequest=false`, so `inFlight.decrementUsed()` is never called to undo the `acquireNow()` increment.

Reproducer test:

```java
@Test
public void testAcceptNullRequestLeaksPoolInFlight(VertxTestContext ctx) {
   var checkpoint = ctx.checkpoint();

   getClientPool(ctx).onComplete(result -> {
      if (result.failed()) return;
      HttpClientPool client = result.result();

      SharedConnectionPool pool = (SharedConnectionPool) client.next();
      pool.executor().execute(() -> {
         Session session = SessionFactory.forTesting();
         HttpRunData.initForTesting(session);

         // Acquire the only connection so the pool is exhausted
         pool.acquire(false, (HttpConnection firstConn) -> {

            // Put a consumer in the waiting queue that mimics
            // HttpRequestContext.accept() when request==null
            pool.acquire(false, (HttpConnection conn) -> {
               if (conn != null && conn.pool() != null) {
                  conn.cancelAcquire();
                  conn.pool().release(conn, true, false);
               }

               ctx.verify(() -> {
                  assertThat(pool.inFlightCount())
                        .as("pool inFlight should be 0 — cancelAcquire undoes conn-level but not pool-level")
                        .isEqualTo(0);  // FAILS: actual is 1
               });
               checkpoint.flag();
            });

            // Release firstConn to trigger pulse() which pops the waiting consumer
            firstConn.cancelAcquire();
            pool.release(firstConn, true, true);
            pool.pulse();
         });
      });
   });
}
```

Result: `expected: 0 but was: 1`

Fix: add `inFlight.decrementUsed()` in the null-request path of `accept()`, same as the [orphaned-acquisition drain](https://github.com/Hyperfoil/Hyperfoil/pull/816/files#diff-2a969a23f7a3ced3e4e9b44e2e6f2e88a6b02dbf91b9b34bcc7bc998d19acbb9R275-R278).

diegolovison:
@franz1981 ready for review


**A quick note on the test tweak and a minor edge case it highlighted:**

I had to add `firstConn.attach(pool);` in the test because our mock doesn't call `request.send()` (which is normally where a connection gets linked to its pool).

Interestingly, this actually exposes a non-obvious edge case in the production code: if a **brand new** connection (which hasn't been attached to a pool yet) hits this `null` request abort *before* `send()` is called, `connection.pool()` will evaluate to null. This means it would bypass the `cancelAcquire` block entirely.

This patch successfully fixes the leak for reused connections (which covers the vast majority of cases), but it suggests that in a future PR, we might want to move `connection.attach(pool)` upstream into the connection creation logic itself, rather than waiting for `HttpRequest.send()`, so connections always know their pool from birth."

franz1981 (approved):
Fix is correct. 138/138 http tests pass. The `cancelAcquire(HttpConnection)` method cleanly consolidates `conn.cancelAcquire()` + `inFlight.decrementUsed()` + `release(conn, true, false)` and is used from both the drain path and `HttpRequestContext.accept()`.

Agree on the brand-new-connection edge case — `connection.pool()` is null until `HttpRequest.send()` calls `connection.attach(pool)`, so a never-used connection would skip the `cancelAcquire` block. Fine to address in a follow-up (moving `attach(pool)` upstream into connection creation).

LGTM.

franz1981:
Well done @diegolovison ! this was a quite deep rabbit hole!

diegolovison:
@franz1981 thanks for the review and for pushing this forward!

PR #823, opened 2026-06-05
Add thread-safety guard to SessionImpl.proceed

diegolovison:
- fix(core): add thread-safety guard to SessionImpl.proceed()
- Include the BufferedReader in the try-with-resources statement together

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/session/SessionImpl.java
+++ b/core/src/main/java/io/hyperfoil/core/session/SessionImpl.java
@@ -453,6 +453,7 @@ class SessionImpl implements Session {
 
    @Override
    public void proceed() {
+      assert executor.inEventLoop();
       if (!scheduled) {
          scheduled = true;
          executor.execute(runTask);
--- a/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
@@ -1,5 +1,6 @@
 package io.hyperfoil.core.impl;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -17,6 +18,8 @@ import io.hyperfoil.api.session.PhaseChangeHandler;
 import io.hyperfoil.api.session.PhaseInstance;
 import io.hyperfoil.api.session.Session;
 import io.netty.util.concurrent.EventExecutorGroup;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
 
 public abstract class PhaseInstanceImpl implements PhaseInstance {
    protected static final Logger log = LogManager.getLogger(PhaseInstanceImpl.class);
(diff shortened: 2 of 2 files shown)


core/src/main/java/io/hyperfoil/core/session/SessionImpl.java
```diff
@@ -453,6 +453,12 @@ private SequenceInstance startSequence(Sequence sequence, boolean forceSameIndex
 
    @Override
    public void proceed() {
+      if (!executor.inEventLoop()) {
+         // Not on event loop - dispatch to it
+         executor.execute(this::proceed);
+         return;
+      }
+      assert executor.inEventLoop();
```
franz1981, line 456: this is redundant as you check it right at runtime

franz1981 (changes requested):

core/src/main/java/io/hyperfoil/core/session/SessionImpl.java
```diff
@@ -453,6 +453,12 @@ private SequenceInstance startSequence(Sequence sequence, boolean forceSameIndex
 
    @Override
    public void proceed() {
+      if (!executor.inEventLoop()) {
+         // Not on event loop - dispatch to it
+         executor.execute(this::proceed);
```
franz1981: this is a bit weird TBH, are you sure about this? 
The caller of `proceed` expect things to be "synchronous" here - but this one is making such call async with a change which is not clear when it would happen

diegolovison:
@franz1981 please review again

franz1981:
The latest version looks correct!

One observation: the proceedPhaseComplete.onComplete() handler only logs: it doesn't retry or propagate the failure. Is that intentional? 
If a proceed() throws, the promise still completes (via finally), so the result.failed() branch would only fire if the promise itself fails, which it won't with this code. 
The error log is effectively dead code.

diegolovison:
@franz1981 Done

Changed to:
```java
  try {
     session.proceed();
     promise.complete();
  } catch (Throwable t) {
     promise.fail(t);
  }
```

franz1981 (approved):
Nice one!

One question: the Future.all(proceedFutures) result is logged but not acted on. 
Should tryTerminate() call setTerminated() once all proceed futures complete (if activeSessions has reached zero by then)?

diegolovison:
@franz1981, what do you think about the following log messages?
If you agree, I will change and merge.

```
Index: core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java b/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java
--- a/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java	(revision b6dd66e20fdb2835d3d685a92f2efc7442c1ce54)
+++ b/core/src/main/java/io/hyperfoil/core/impl/PhaseInstanceImpl.java	(date 1782399117936)
@@ -168,9 +168,9 @@
 
          proceedPhaseComplete.onComplete(result -> {
             if (result.succeeded()) {
-               log.debug("All sessions terminated for phase '{}'.", def.name);
+               log.debug("Successfully triggered proceed() on background sessions for phase '{}'.", def.name);
             } else {
-               log.error("Failed to terminate all sessions for phase '{}'", def.name, result.cause());
+               log.error("Failed to call proceed() on asynchronously handled sessions for phase '{}'", def.name, result.cause());
             }
          });
       }
```

franz1981 (approved):
LGTM well done @diegolovison !

PR #824, opened 2026-06-08
Bump Vert.x to 5.1.1

diegolovison:
Bump Vert.x to 5.1.1

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,3 +14,8 @@ updates:
     open-pull-requests-limit: 10
     labels:
       - dependency
+    ignore:
+      - dependency-name: "org.infinispan:*"
+      - dependency-name: "com.github.ben-manes.caffeine:caffeine"
+      - dependency-name: "io.reactivex.rxjava3:rxjava"
+      - dependency-name: "org.wildfly.common:wildfly-common"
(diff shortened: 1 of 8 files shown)


stalep:
Seems like a clean followup version bump from vert.x 5.0.
Do we have good enough tests to verify the Inifispan upgrade?
Also, the explicit wildfly-common dependency, can't we resolve that transitively?

We need to make sure there are no regressions with this upgrade. That's vital before merging. Otherwise looks good!

diegolovison:
> Do we have good enough tests to verify the Inifispan upgrade?

I believe that the current test suite is good enough to catch regressions. At least, it failed for every minor/major upgrade of Vert.x and Infinispan.

> Also, the explicit wildfly-common dependency, can't we resolve that transitively?

https://github.com/vert-x3/vertx-infinispan/issues/154 https://github.com/vert-x3/vertx-infinispan/commit/04241a79c745ac22fedb39fe3569c3d21fb72dcf
We will need to wait for the next release.

franz1981:
@diegolovison the whole move to the new vertx 5 version requires assessing we have no perf regressions either :pray: 
especially in term of allocation rate - or we have to send fixes to vertx if we spot any.
And that means, regardless, this is not part of a minor, but we have to release a major likely wdyt?

franz1981 (approved):

PR #825, opened 2026-06-08
Remove developer section from pom.xml

diegolovison:
Tracking individual contributors in the file is redundant since this information is already maintained in the Git history.


franz1981:
@stalep wdyt? I'm fine w it tbh

franz1981 (commented):
See my comment

diegolovison:
@stalep any objection?

PR #843, opened 2026-07-01
Disable auto-formatting locally, enforce in CI

diegolovison:
Fix concurrent file modification issues between LLMs and Maven commands.


diegolovison:
@franz1981 @stalep, what do you think?
The LLM is losing context and making wrong edits because it loses context.

PR #848, opened 2026-07-06
[#847] Support clustered benchmarks via CLI

### LINKED ISSUE #847: Support clustered benchmarks via CLI
**Is your feature request related to a problem? Please describe.**
The standalone CLI commands always start a non-clustered controller. Benchmarks defining `agents:` are rejected because. Running distributed benchmarks currently requires deploying the full Hyperfoil server distribution.

**Describe the solution you'd like**
A `--clustered` flag on the standalone CLI that starts a clustered Vert.x instance, enabling SSH-deployed agents on remote machines. The flag would be intercepted in `LoadAndRun.main()` and threaded through `StartLocal` to `LocalController.Factory.start()`, which would use `Hyperfoil.clusteredVertx(true)` instead of `Vertx.vertx()`.

**Additional context**
This also allows us to utilize the JBang command to submit clustered benchmarks: `jbang --fresh run@. --clustered sample-benchmark.yml` where `sample-benchmark.yml` defines multiple remote agents. The local node where the command is run serves as the controller.

jabolina:
* Include the `--clustered` flag when running a clustered benchmark with the CLI.

Closes #847

## Changes proposed

Included a `--clustered` flag to pass along to indicate whether it is running a benchmark as clustered or in-VM. Without the flag, we would need to parse the provided benchmark file to identify whether it has the agents. I think just passing the flag is simpler. It moves the complexity to the user, though.

### DIFF
--- a/api/src/main/java/io/hyperfoil/internal/Controller.java
+++ b/api/src/main/java/io/hyperfoil/internal/Controller.java
@@ -23,6 +23,6 @@ public interface Controller {
    void stop();
 
    interface Factory {
-      Controller start(Path rootDir);
+      Controller start(Path rootDir, boolean clustered);
    }
 }
--- a/cli/src/main/java/io/hyperfoil/cli/commands/BaseStandaloneCommand.java
+++ b/cli/src/main/java/io/hyperfoil/cli/commands/BaseStandaloneCommand.java
@@ -63,7 +63,7 @@ public abstract class BaseStandaloneCommand {
          cr = runtime.build();
          try {
             // start the local in-vm controller server
-            cr.executeCommand("start-local --quiet");
+            cr.executeCommand(startLocalCommand());
             // As -H option could contain a whitespace we have to either escape the space or quote the argument.
             // However, quoting would not work well if the argument contains a quote.
             String optionsCollected = Stream.of(args).map(arg -> arg.replaceAll(" ", "\\\\ ")).collect(Collectors.joining(" "));
@@ -89,6 +89,10 @@ public abstract class BaseStandaloneCommand {
       }
 
       return result == null ? CommandResult.FAILURE.getResultValue() : result.getResultValue();
+   }
+
+   protected String startLocalCommand() {
+      return "start-local --quiet";
    }
 
    /**
--- a/cli/src/main/java/io/hyperfoil/cli/commands/LoadAndRun.java
+++ b/cli/src/main/java/io/hyperfoil/cli/commands/LoadAndRun.java
@@ -1,6 +1,8 @@
 package io.hyperfoil.cli.commands;
 
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Stream;
 
 import org.aesh.command.Command;
 import org.aesh.command.CommandDefinition;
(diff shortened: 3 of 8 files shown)


diegolovison:
Hi @jabolina Please confirm the following:

### The Scenario: Generating massive load

You are testing a new microservice and need to hit it with 100,000 requests per second. A single machine cannot generate this much traffic due to CPU and network limits. You need to distribute the load generation across three remote servers (Agents).

You write your Hyperfoil benchmark file (`heavy-load.yml`) and define your remote agents at the top:

```yaml
name: heavy-load-test
agents:
  agent1: host=192.168.1.10
  agent2: host=192.168.1.11
  agent3: host=192.168.1.12
# ... rest of HTTP benchmark ...

```

---

### The Problem (Before this PR)

If you tried to run this using the standalone CLI command (`run heavy-load.yml`), Hyperfoil would **reject** the benchmark and throw an error. Why? Because the standalone CLI always started a local, isolated (non-clustered) controller that didn't know how to talk to remote machines over a network.

To actually run this distributed test, you had to do a lot of manual infrastructure work:

1. Download the full Hyperfoil server distribution.
2. Start the Hyperfoil Controller server in the background as a standalone Java process.
3. Open the Hyperfoil CLI and connect to that background server.
4. Upload your `heavy-load.yml` file to the server.
5. Trigger the run.

This was tedious for local development and made integrating distributed benchmarks into automated CI/CD pipelines (like GitHub Actions) clunky.

---

### The Solution (After this PR)

With the new `--clustered` flag, you can skip all the manual server setup. You just run a single command from your local laptop or CI runner:

```bash
jbang run@. --clustered heavy-load.yml

```

### "Start agents and what more?" (What happens under the hood)

When you add that `--clustered` flag, here is the exact sequence of events the PR enables seamlessly in the background:

1. **Starts an Embedded Controller:** Instead of spinning up a standalone JVM for the server, the CLI boots up an embedded Vert.x Controller directly inside the CLI process, but critically, it boots it in **clustered mode** (capable of network discovery).
2. **Deploys Remote Agents via SSH:** The controller reads your YAML, sees `agent1`, `agent2`, and `agent3`, and uses SSH to automatically connect to those remote machines.
3. **Bootstraps the Cluster:** It pushes the Hyperfoil agent code to those remote machines and starts them. Because the local controller is running in `--clustered` mode, the agents successfully discover the controller on the network and join the Vert.x cluster.
4. **Orchestrates the Load:** The local CLI controller slices up the benchmark (e.g., giving 33,333 req/sec to each agent) and sends the marching orders over the clustered event bus.
5. **Aggregates Real-Time Metrics:** As the three remote agents blast the target server with traffic, they stream their live metrics back to your local CLI over the cluster network. You see the unified results right in your terminal.
6. **Automatic Teardown:** Once the benchmark finishes, the local controller signals the remote agents to shut down, closes the SSH connections, terminates the Vert.x cluster, and exits the CLI.

**Summary:** It turns a complex, multi-machine distributed systems deployment into a single, automated, one-line command.

jabolina:
Diego, that's right. The `--clustered` option utilises the existing infra for running the benchmark clustered. The process running the CLI becomes the controller, and it'll deploy, run, and collect results.

diegolovison:
Thanks @jabolina

PR #851, opened 2026-07-13
hyperfoil-comparison.sh to automate the setup, build, and execution of Hyperfoil wrk and wrk2 benchmarks

diegolovison:
hyperfoil-comparison.sh to automate the setup, build, and execution of Hyperfoil wrk and wrk2 benchmarks

### DIFF
--- a/benchmarks/hyperfoil-comparison.sh
+++ b/benchmarks/hyperfoil-comparison.sh
@@ -1,0 +1,143 @@
+#!/bin/bash
+set -eo pipefail
+
+# ==========================================
+# Default Configuration (States)
+# ==========================================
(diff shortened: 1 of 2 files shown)


franz1981 (changes requested):
Wouldn't be better if we use a qDup script as well?
Ideally we could setup a serie of different benchmarks and onboard huperfoil on h5m bringing per pr benchmarks on the lab.
This would allow us to improve the whole chain and keep the huperfoil performance under control.

For the first time we likely don't want the server to be the bottleneck ^^ 
In addition to this we need to "pin" the server to be stable as we don't want it to impact our load gen numbers

PR #858, opened 2026-07-27
Add wrk2 test with big timeout

diegolovison:
Add wrk2 test with big timeout

### DIFF
--- a/core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsCollector.java
+++ b/core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsCollector.java
@@ -29,8 +29,8 @@ public class StatisticsCollector implements Consumer<SessionStatistics> {
 
    @Override
    public void accept(SessionStatistics statistics) {
-      if (log.isDebugEnabled()) {
-         log.debug("{}: Received statistics", statistics);
+      if (log.isTraceEnabled()) {
+         log.trace("Received statistics for: {}", statistics);
       }
       for (int i = 0; i < statistics.size(); ++i) {
          int phaseAndStepId = (statistics.phase(i).id() << 16) + statistics.step(i);
@@ -39,7 +39,7 @@ public class StatisticsCollector implements Consumer<SessionStatistics> {
          if (metricMap == null) {
             metricMap = new HashMap<>();
             if (log.isDebugEnabled()) {
-               log.debug("{}: Aggregated metric statistics for phaseAndStepId={}", statistics, phaseAndStepId);
+               log.debug("Aggregated metric statistics for phaseAndStepId={}", phaseAndStepId);
             }
             aggregated.put(phaseAndStepId, metricMap);
          }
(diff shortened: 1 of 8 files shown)


test-suite/src/test/resources/log4j2.xml
```diff
@@ -21,7 +21,7 @@
         <Root level="INFO">
             <AppenderRef ref="STDOUT" level="INFO"/>
 
-            <AppenderRef ref="FILE" level="TRACE"/>
+            <AppenderRef ref="FILE" level="INFO"/>
```
franz1981: why?

diegolovison: Forgot to rollback.

core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsCollector.java
```diff
@@ -39,7 +39,7 @@ public void accept(SessionStatistics statistics) {
          if (metricMap == null) {
             metricMap = new HashMap<>();
             if (log.isDebugEnabled()) {
-               log.debug("{}: Aggregated metric statistics for phaseAndStepId={}", statistics, phaseAndStepId);
+               log.debug("Aggregated metric statistics for phaseAndStepId={} with values={}", phaseAndStepId, metricMap);
```
stalep: I think metricMap will always be empty here?

http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
```diff
@@ -96,8 +96,8 @@ private HttpConnection acquireNow(boolean exclusiveConnection) {
                   usedConnections.incrementUsed();
                }
                connection.onAcquire();
-               if (log.isDebugEnabled()) {
-                  log.debug("acquireNow: acquired {} to {}, inFlight={} (after), connection.inFlight={}",
+               if (log.isTraceEnabled()) {
```
stalep: We have a trace field initiated on line 36 we could use instead. There are other instances below with the same pattern.

http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
stalep: Should we change this as well to log.trace? Similar as above?

stalep (commented):
Some minor comments

http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
diegolovison: Debug level is good. I added `isDebugEnabled` because `ArrayDeque::size`

http/src/main/java/io/hyperfoil/http/connection/SharedConnectionPool.java
diegolovison: Yes. Done for the others as well

core/src/main/java/io/hyperfoil/core/impl/statistics/StatisticsCollector.java
diegolovison: Done

PR #870, opened 2026-08-14
fix(core): calculate throughput using request fired time instead of scheduled time

### LINKED ISSUE #860: wrk2 report wrong Req/Sec
## Description

Using the branch https://github.com/diegolovison/Hyperfoil/tree/req_sec we can troubleshoot the following behavior:
- when comparing wrk2 with wrk we have more `Req/Sec`. That behavior must not be true. See the following output:

**Server**
```
$ taskset -c 2-5 jbang /tmp/server/MockHttpServer.java --port 8080 --think-time 50 --threads 4

Mock HTTP Server started on port 8080 with 50000000ns think time using 4 thread(s)
```

**Hyperfoil**
```
$ taskset -c 12-15 java -jar /home/dlovison/github/Hyperfoil/Hyperfoil/clustering/target/wrk.jar --timeout 2s --threads 1 --connections 5 --duration 30s http://localhost:8080/fruits

Running 30s test @ http://localhost:8080/fruits
  1 threads and 5 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    50.86ms  202.24μs  52.69ms    0.00%
    Req/Sec    98.00      2.45   100.00     60.00
  2940 requests in 30.011s,   2.06MB read
Requests/sec: 97.96
Transfer/sec:  70.39kB

$ taskset -c 12-15 java -jar /home/dlovison/github/Hyperfoil/Hyperfoil/clustering/target/wrk2.jar --timeout 2s --threads 1 --connections 5 --duration 30s --rate 50000 http://localhost:8080/fruits

Running 30s test @ http://localhost:8080/fruits
  1 threads and 5 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    15.04s     8.63s   29.93s    57.44%
    Req/Sec   2925.00      0.00  2925.00    100.00
  2925 requests in 30.963s,   2.05MB read
Requests/sec: 94.47
Transfer/sec:  67.88kB

$ taskset -c 12-15 java -jar /home/dlovison/github/Hyperfoil/Hyperfoil/clustering/target/wrk2.jar --timeout 100s --threads 1 --connections 5 --duration 30s --rate 50000 http://localhost:8080/fruits

Running 30s test @ http://localhost:8080/fruits
  1 threads and 5 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    14.96s     8.61s   29.80s    57.41%
    Req/Sec   2935.00      0.00  2935.00    100.00
  2935 requests in 30.395s,   2.06MB read
Requests/sec: 96.56
Transfer/sec:  69.39kB
```

The number of requests are accurate the problem is that how Hyperfoil is doing the calculation
```java
DoubleSummaryStatistics requestsStats = series.stream().mapToDouble(s -> s.requestCount).summaryStatistics();
```
then
```
requestsStats.getAverage()
```
because we have only **one** bucket it will divide the number of requests per bucket.

---

## How can I see this behavior?

- Start the server
- Execute the test
```
rm -fR /tmp/hyperfoil/ && mvn test -Pbenchmark -pl clustering,cli,api,core,http,test-suite/ -Dtest=WrkTest#testWrk2LocalServer -Dio.hyperfoil.controller.log.level=debug -Dsurefire.failIfNoSpecifiedTests=false
```
- See the wrk2 command output
```
$ cat /home/dlovison/github/Hyperfoil/Hyperfoil/test-suite/target/surefire-reports/io.hyperfoil.benchmark.standalone.WrkTest-output.txt

/tmp/junit-520444266331087167/81b577c5-f7c2-4211-b2c3-868a75d27707.html
Running 30s test @ localhost:8080/fruits
  1 threads and 5 connections
Report written to /tmp/junit-520444266331087167/81b577c5-f7c2-4211-b2c3-868a75d27707.html
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    14.97s     8.62s   29.93s    57.31%
    Req/Sec   2975.00      0.00  2975.00    100.00
  2975 requests in 30.327s,   2.09MB read
Requests/sec: 98.10
Transfer/sec:  70.49kB
-----------------
calibration 790 785
test 2975 2970
-----------------
```
As you can see there is 2975 requests that is reported accurate in `2975 requests in 30.327s` and the problem becomes `Req/Sec   2975.00` that I already told the reason above.
- Execute the report to understand how Hyperfoil behaves. The following report shows response counters, that is when we track statistics. Let's consider for now that it is OK to have a high number of requests when compared with responses. The original wrk2 tracks response only. See https://github.com/Hyperfoil/Hyperfoil/issues/862 . As you can see the response number is 785 for `calibration` phase and 2970 for `test` phase. 
```
$ ./show_stats.sh 

TIMESTAMP            | REQUESTS CREATED | ACTIVE STATS GETS    | RIGHT STATS CREATION
--------------------------------------------------------------------------------------------------
2026-07-28 12:43:24  | 32640            | 785                  | 60                  
2026-07-28 12:43:25  | 50000            | 0                    | 95                  
2026-07-28 12:43:26  | 49999            | 0                    | 100                 
2026-07-28 12:43:27  | 50001            | 0                    | 100                 
2026-07-28 12:43:28  | 49999            | 0                    | 100                 
2026-07-28 12:43:29  | 50002            | 0                    | 100                 
2026-07-28 12:43:30  | 17269            | 0                    | 95                  
2026-07-28 12:43:31  | 0                | 0                    | 100                 
2026-07-28 12:43:32  | 24940            | 2970                 | 75                  
2026-07-28 12:43:33  | 50001            | 0                    | 100                 
2026-07-28 12:43:34  | 49999            | 0                    | 100                 
2026-07-28 12:43:35  | 50001            | 0                    | 100                 
2026-07-28 12:43:36  | 49999            | 0                    | 95                  
2026-07-28 12:43:37  | 50000            | 0                    | 100                 
2026-07-28 12:43:38  | 50000            | 0                    | 100                 
2026-07-28 12:43:39  | 50001            | 0                    | 100                 
2026-07-28 12:43:40  | 49999            | 0                    | 100                 
2026-07-28 12:43:41  | 50001            | 0                    | 100                 
2026-07-28 12:43:42  | 49999            | 0                    | 100                 
2026-07-28 12:43:43  | 49999            | 0                    | 95                  
2026-07-28 12:43:44  | 50002            | 0                    | 100                 
2026-07-28 12:43:45  | 49998            | 0                    | 100                 
2026-07-28 12:43:46  | 50001            | 0                    | 100                 
2026-07-28 12:43:47  | 26590            | 0                    | 100                 
2026-07-28 12:43:48  | 100              | 0                    | 100                 
2026-07-28 12:43:49  | 100              | 0                    | 100                 
2026-07-28 12:43:50  | 100              | 0                    | 100                 
2026-07-28 12:43:51  | 95               | 0                    | 95                  
2026-07-28 12:43:52  | 100              | 0                    | 100                 
2026-07-28 12:43:53  | 100              | 0                    | 100                 
2026-07-28 12:43:54  | 100              | 0                    | 100                 
2026-07-28 12:43:55  | 100              | 0                    | 100                 
2026-07-28 12:43:56  | 100              | 0                    | 100                 
2026-07-28 12:43:57  | 95               | 0                    | 95                  
2026-07-28 12:43:58  | 100              | 0                    | 100                 
2026-07-28 12:43:59  | 100              | 0                    | 100                 
2026-07-28 12:44:00  | 100              | 0                    | 100                 
2026-07-28 12:44:01  | 100              | 0                    | 100                 
2026-07-28 12:44:02  | 50               | 0                    | 50                  
--------------------------------------------------------------------------------------------------
TOTAL                | 1052880          | 3755                 | 3755    
```
This is also affecting the report open `/tmp/junit-520444266331087167/81b577c5-f7c2-4211-b2c3-868a75d27707.html` in your browser. You see the following:

Image

Only one second being reported that is not accurate. We have responses as you can see in "RIGHT STATS CREATION" column above.

---

## Summary
- Wrong `Req/Sec` reported by wrk2 due the number of requests being created
- Wrong report

## Out of the scope
https://github.com/Hyperfoil/Hyperfoil/issues/861
https://github.com/Hyperfoil/Hyperfoil/issues/862

diegolovison:
## Description

Resolves an issue where wrk2 mode reported unrealistically high Req/Sec during server saturation. By bucketing statistics using the moment the request is actually fired (`firedTimestampMillis`) rather than its intended schedule, metrics correctly span the full test duration instead of collapsing.

Fixes: https://github.com/Hyperfoil/Hyperfoil/issues/860

## The Solution
The patch switches Hyperfoil's statistic bucketing strategy from Intended Time to Fired Time (Option B from the issue discussion).

#### 1. Tracking Actual Fired Timestamps (StartTimeSource & Implementations)
- Added getFiredTimestampMillis() and setFiredTimestampMillis() methods to the StartTimeSource interface.
- Updated request implementations (Request.java, HttpRequest.java, HotRodRequestStep.java, StopwatchBeginStep.java, DelaySessionStartStep.java) to capture System.currentTimeMillis() at the exact moment the request is sent/fired.

#### 2. Changing the Bucketing Calculation (Statistics.java)

- Modified Statistics#active(...) to compute the sampling array index using the fired timestamp instead of the intended start timestamp:

```java
// Old:
long startTimestampMillis = source.getStartTimestampMillis(session);

// New:
long recordTimestampMillis = source.getFiredTimestampMillis(session);
int index = (int) ((recordTimestampMillis - startTimestamp) / SAMPLING_PERIOD_MILLIS);
```

- This ensures requests fall into time buckets corresponding to when Hyperfoil actually generated the traffic.

#### 3. Exposing Time-Series Data in CLI (WrkAbstract.java)
- Updated WrkCommandResult to include the metric time series list (List<StatisticsSummary> series). This enables CLI tools and integration tests to verify the distribution of stats across time buckets.

#### 4. Regression Test (WrkTest.java)
- Added testWrk2ReqSec: Runs a 15-second test (-d 15s) against a /50ms endpoint and asserts that series.size() == 15. This verifies that statistics are now distributed across 15 individual 1-second buckets instead of collapsing.

## Report
Before this change the report was (1 bucket per phase)
image

After this change the report will be ( distributed buckets per fire time)
image

### DIFF
--- a/api/src/main/java/io/hyperfoil/api/config/StartTimeSource.java
+++ b/api/src/main/java/io/hyperfoil/api/config/StartTimeSource.java
@@ -21,6 +21,7 @@ public interface StartTimeSource {
     * @param target a pre-allocated {@code long[]} buffer to store [millis, nanos]
     */
    default void createStartTimestamp(Session session, boolean useSessionStartTime, long[] target) {
+      assert target.length == 2;
       if (useSessionStartTime) {
          long sessionStartTime = session.scheduledStartTimestamp();
          long sessionStartNanoTime = session.scheduledStartNanoTime();
@@ -38,4 +39,8 @@ public interface StartTimeSource {
    long getStartTimestampMillis(Session session);
 
    long getStartTimestampNanos(Session session);
+
+   long getFiredTimestampMillis(Session session);
+
+   void setFiredTimestampMillis(Session session);
 }
--- a/api/src/main/java/io/hyperfoil/api/connection/Request.java
+++ b/api/src/main/java/io/hyperfoil/api/connection/Request.java
@@ -26,6 +26,7 @@ public abstract class Request implements Callable<Void>, GenericFutureListener<F
    private final long[] timestamps = new long[2];
    private long startTimestampMillis;
    private long startTimestampNanos;
+   private long firedTimestampMillis;
    private SequenceInstance sequence;
    private SequenceInstance completionSequence;
    private Statistics statistics;
@@ -145,6 +146,15 @@ public abstract class Request implements Callable<Void>, GenericFutureListener<F
 
    public long startTimestampNanos() {
       return startTimestampNanos;
+   }
+
+   public long firedTimestampMillis() {
+      return this.firedTimestampMillis;
+   }
+
+   @Override
+   public void setFiredTimestampMillis(Session session) {
+      this.firedTimestampMillis = System.currentTimeMillis();
    }
 
    public void setTimeout(long timeout, TimeUnit timeUnit) {
--- a/api/src/main/java/io/hyperfoil/api/statistics/Statistics.java
+++ b/api/src/main/java/io/hyperfoil/api/statistics/Statistics.java
@@ -132,7 +132,8 @@ public class Statistics {
    public void incrementBlockedTime(StartTimeSource source, long blockedTime, Session session) {
       long criticalValueAtEnter = recordingPhaser.writerCriticalSectionEnter();
       try {
-         StatisticsSnapshot active = active(source, session);
+         // it can be or not fired. track always as intended time
+         StatisticsSnapshot active = active(source.getStartTimestampMillis(session));
          active.blockedTime += blockedTime;
       } finally {
          recordingPhaser.writerCriticalSectionExit(criticalValueAtEnter);
@@ -259,15 +260,19 @@ public class Statistics {
    }
 
    private StatisticsSnapshot active(StartTimeSource source, Session session) {
-      long startTimestampMillis = source.getStartTimestampMillis(session);
-      assert startTimestampMillis > 0;
-      int index = (int) ((startTimestampMillis - startTimestamp) / SAMPLING_PERIOD_MILLIS);
+      long recordTimestampMillis = source.getFiredTimestampMillis(session);
+      assert recordTimestampMillis > 0;
+      return this.active(recordTimestampMillis);
+   }
+
+   private StatisticsSnapshot active(long recordTimestampMillis) {
+      int index = (int) ((recordTimestampMillis - startTimestamp) / SAMPLING_PERIOD_MILLIS);
       AtomicReferenceArray<StatisticsSnapshot> active = this.active;
       if (index >= active.length()) {
          active = resizeArray(active, index);
          this.active = active;
       } else if (index < 0) {
-         log.error("Record start timestamp {} predates statistics start {}", startTimestampMillis,
+         log.error("Record start timestamp {} predates statistics start {}", recordTimestampMillis,
                startTimestamp);
          index = 0;
       }
--- a/cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
+++ b/cli/src/main/java/io/hyperfoil/cli/commands/WrkAbstract.java
@@ -173,7 +173,7 @@ public abstract class WrkAbstract extends BaseStandaloneCommand {
          boolean result = awaitBenchmarkResult(run, invocation);
 
          RequestStatisticsResponse total = run.statsTotal();
-         wrkCommandResult = new WrkCommandResult(run, total);
+         wrkCommandResult = new WrkCommandResult(run, total, run.series("test", 0, "request"));
 
          if (result) {
             if (output != null && !output.isBlank()) {
@@ -249,7 +249,7 @@ public abstract class WrkAbstract extends BaseStandaloneCommand {
       protected abstract PhaseBuilder<?> phaseConfig(PhaseBuilder.Catalog catalog, WrkScenario.PhaseType phaseType, long durationMs);
 
       private void printStats(StatisticsSummary stats, AbstractHistogram histogram, List<StatisticsSummary> series,
-            CommandInvocation invocation) {
+                              CommandInvocation invocation) {
          TransferSizeRecorder.Stats transferStats = (TransferSizeRecorder.Stats) stats.extensions.get("transfer");
          HttpStats httpStats = HttpStats.get(stats);
          double durationSeconds = (stats.endTime - stats.startTime) / 1000d;
@@ -324,12 +324,14 @@ public abstract class WrkAbstract extends BaseStandaloneCommand {
    }
 
    public static class WrkCommandResult {
-      Client.RunRef run;
-      RequestStatisticsResponse requestStatisticsResponse;
+      final Client.RunRef run;
+      final RequestStatisticsResponse requestStatisticsResponse;
+      final List<StatisticsSummary> series;
 
-      public WrkCommandResult(Client.RunRef run, RequestStatisticsResponse requestStatisticsResponse) {
+      public WrkCommandResult(Client.RunRef run, RequestStatisticsResponse requestStatisticsResponse, List<StatisticsSummary> series) {
          this.run = run;
          this.requestStatisticsResponse = requestStatisticsResponse;
+         this.series = series;
       }
 
       public Client.RunRef getRun() {
@@ -338,6 +340,10 @@ public abstract class WrkAbstract extends BaseStandaloneCommand {
 
       public RequestStatisticsResponse getRequestStatisticsResponse() {
          return requestStatisticsResponse;
+      }
+
+      public List<StatisticsSummary> getSeries() {
+         return series;
       }
    }
 }
--- a/core/src/main/java/io/hyperfoil/core/steps/DelaySessionStartStep.java
+++ b/core/src/main/java/io/hyperfoil/core/steps/DelaySessionStartStep.java
@@ -76,6 +76,7 @@ public class DelaySessionStartStep implements Step, ResourceUtilizer {
       public double period;
       public ScheduledFuture<?> future;
       public PhaseInstance phase;
+      private long firedTimestampMillis;
 
       public long lastStartTime() {
          return startTimeWithOffset + (long) ((iteration - 1) * period);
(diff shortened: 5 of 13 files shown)


stalep:
I have one concern about `DelaySessionStartStep.Holder`:

The PR adds getFiredTimestampMillis() / setFiredTimestampMillis() to the StartTimeSource interface and changes Statistics.active() to use `source.getFiredTimestampMillis(session)` instead of `source.getStartTimestampMillis(session)`. The Holder class implements these new methods, but I don't see any code path that actually calls `setFiredTimestampMillis()` on a Holder instance, so firedTimestampMillis stays at 0.

This matters because `CompensatedResponseRecorder.run()` already passes holder as a StartTimeSource to `statistics.incrementRequests(holder, session)` and `statistics.recordResponse(holder, ..., session)`. After this PR, those calls route through active() which now reads `holder.getFiredTimestampMillis(session)`, returning 0. With -ea this hits the assert recordTimestampMillis > 0; without assertions it computes a negative index, logs an error, and collapses everything into bucket 0.

Would adding `holder.setFiredTimestampMillis(session) in DelaySessionStartStep.invoke()` (before starting the sequences) fix this?

A few smaller items:

1. StopwatchBeginStep sets the field directly (`startTime.firedTimestampMillis = System.currentTimeMillis()`) instead of calling setFiredTimestampMillis(session) like the other call sites. Intentional?
2. WrkTest.java:161 uses assert series != null (Java assert, no-op without -ea) rather than assertNotNull(series). The rest of the test uses JUnit assertions.

diegolovison:
@stalep @franz1981 please review

franz1981:
Just to see if I got it right; the intended start time is "valid" only at 0 hop from the actual source of the tick - right?
This PR is applying this regardless this structural discriminator?

diegolovison:
The "intended start time" represents the theoretical schedule (the 0-hop tick). However, once saturation occurs or connections block, the actual execution drifts significantly from that schedule.

This PR applies the "fired time" universally across all requests **for the purpose of bucketing metrics**, regardless of that structural discriminator.

Here is why this shift was made:

* **The Problem with Intended Time:** Bucketing by the 0-hop intended time causes retroactive metric updates. If a server saturates and blocks connections for 10 seconds, requests scheduled for seconds 1-10 might actually fire in second 11. Under the old strategy, they were retroactively placed in buckets 1-10, masking the bottleneck and artificially inflating the `Req/Sec` to look like it hit the target throughput.
* **The Fired Time Solution:** By universally bucketing based on `firedTimestampMillis` (when the request actually leaves Hyperfoil), the metrics now accurately reflect real generator activity. Past buckets become immutable once the generator moves past that second.
* **The Fallback:** The PR does maintain the intended `startTimestampMillis`, but only as a fallback for bucketing in cases where a request fails or times out *before* it actually fires.

franz1981:
But it still means then a previous fire time on 2 costly N hops while nearly saturated will be recorded as next to the later one when using the intended will kick in for a later one and that doesn't sound good either..

diegolovison:
You are right. This is exactly the 'Con' of Option B that I mapped out in the issue. https://github.com/Hyperfoil/Hyperfoil/issues/860#issuecomment-5144401905

As you can see from the table all implementations has a Pro/Con. We need to decide which one is the less problematic. I decided with option B and started this PR.

Do you agree this trade-off is our best path forward, or should we pivot to a different strategy?

stalep:
Just chiming in here after another look at the code. Nice updates Diego, the issues I raised are all fixed.

The fallback in Statistics.active() is a good safety net for requests that fail before setFiredTimestampMillis gets called. One thought: if a future StartTimeSource implementor forgets to call setFiredTimestampMillis, the fallback silently reverts to intended-time bucketing, which is the original bug. A trace-level log on that path would make it observable without adding noise.

Regarding Franz's question about the N-hop case near saturation: I think this is worth resolving before we go further with code-level items. The tradeoff table in #860 is clear, and Option B does have the known con that blocked connections produce zero-throughput buckets followed by a surge. Whether that's acceptable depends on what the primary consumer of these metrics expects. If the goal is CI gating (like the micronaut-http-benchmarks use case), fired time gives stable, non-retroactive buckets. If the goal is accurate per-second throughput accounting, the surge effect could mislead.

Pending that design decision, a few implementation notes for when we move forward:
1. The `firedTimestampMillis` field + getter + setter is duplicated identically in `Request`, `DelaySessionStartStep.Holder`, `StopwatchBeginStep.StartTime`, and `HotRodRequestStep`. Every new `StartTimeSource` implementor has to copy the same snippet. A default method on the interface could collapse this.
2. `setFiredTimestampMillis(Session session)` ignores the session parameter in three of four implementations. It only needs `Session` for `HotRodRequestStep`'s resource lookup.
3. `testWrk2ReqSec` checks the bucket count but doesn't verify that the sum of requests across buckets equals the total. Worth adding to confirm nothing gets dropped during rebucketing.

diegolovison:
@franz1981 @stalep 

How about if we reach a consensus on the bucketing strategy (Option A, B or C -> https://github.com/Hyperfoil/Hyperfoil/issues/860#issuecomment-5144401905)  before I push any other change ?

I'm happy to pivot if we decide another route fits the primary consumers better.

stalep:
I lean toward Option B being the right tradeoff for Hyperfoil's primary use cases, but Franz's concern about multi-hop requests near saturation deserves a clear answer. @franz1981 what do you think?

franz1981:
**Option B**, once blocked time stops being dropped.

I measured it with a 2-hop open-model test (`compensateInternalLatency` on), a server that is slow for a known window, and every stats write traced, then re-bucketed as master / B / C:
- **Multi-hop mixing:** real, but master already has it (worst bucket spans 5.9s of intended starts vs 5.7s with B).
- **Throughput:** B per-second counts match the server's arrivals exactly; master shows a flat 200/s through the stall.
- **Onset:** B shows the slowdown in the right second; C shows it 2s late (the service time).
- **Late writes to past buckets:** master 2800 (up to 5s), B 74 (up to 3s).

**Blocker:** `incrementBlockedTime` still keys on start time, so some buckets hold only `blockedTime`, and `publish()` drops them (`isEmpty()` ignores it). Only 1137s of 3439s blocked time got published; master loses none. Please key it like the counts and include `blockedTime` in `isEmpty()`. Also reset `firedTimestampMillis` in `Request.start()`.

diegolovison:
@franz1981 all comments were addressed
Please review again https://github.com/Hyperfoil/Hyperfoil/pull/870/changes/b284504691c44f5e620abb14f24e172c534aa017

PR #872, opened 2026-08-25
Add declarative response body checks

### LINKED ISSUE #3: Upstream declarative Hyperfoil response-body validation
## Goal

Add stock, declarative Hyperfoil YAML support for validating complete HTTP response bodies so endpoint smoke benchmarks do not require a project-specific `Processor` extension.

The required matching modes are:

- Exact body equality.
- Full regular-expression match.
- Structural JSON equality.

## Motivation

Hyperfoil already provides the required processing infrastructure:

- `httpRequest.handler.status` for status validation.
- `httpRequest.handler.body` processors with fragmented-buffer lifecycle support.
- Request invalidation and `stopOnInvalid`.
- Streaming JSON parsing.

However, the existing `SearchValidator` only counts substring occurrences. It cannot express complete-body equality, `Matcher.matches()` semantics, or structural JSON equality.

The benchmark suite currently models these modes as `EQUAL`, `REGEX`, and `JSON`. A stock Hyperfoil equivalent would let Nix generate self-contained smoke YAML with no extension JAR or ServiceLoader registration.

## Proposed upstream work

Introduce a built-in declarative body validator, for example:

```yaml
httpRequest:
  GET: /endpoint
  handler:
    body:
      check:
        equalTo: '{"result":1}'
```

Alternative strategies under the same `check` builder:

```yaml
check:
  regex: '^prefix-.*-suffix$'
```

```yaml
check:
  json: '{"result":1}'
```

The builder should require exactly one strategy.

Expected semantics:

- `equalTo`: complete body and length equality, with no ignored prefix/suffix.
- `regex`: full-body `Matcher.matches()`, not substring `find()`.
- `json`: structural equality; ignore insignificant whitespace and object-key order, preserve array order, and document numeric comparison semantics.
- All modes must work when the HTTP body is delivered in multiple fragments.
- Mismatches must mark the request invalid and integrate with `stopOnInvalid` and strict CLI failure reporting.

## Upstream deliverables

- Built-in processor and YAML builder.
- Generated JSON schema support.
- Reference documentation and examples.
- Tests for fragmented bodies and each matching mode.
- Tests for extra/missing bytes, regex substring-only mismatch, malformed JSON, object-key order, array order, and invalidation behavior.

## Repository integration

After an upstream release includes the validator:

- Update the pinned Hyperfoil version.
- Generate stock `handler.body.check` YAML from the existing Nix `responseMatchingMode` data.
- Remove the custom smoke response-validator class and extension packaging.
- Continue using Hyperfoil status validation for readiness checks and allow configured non-2xx benchmark responses where intended.

## Acceptance criteria

- Nix-generated smoke YAML represents all current `EQUAL`, `REGEX`, and `JSON` modes using stock Hyperfoil features.
- Mismatched responses reliably invalidate the smoke run.
- Fragmented responses produce the same result as contiguous responses.
- `micronaut-http-benchmarks` no longer packages a custom response processor or ServiceLoader metadata.

yawkat:
## Fixes Issue

Adds the upstream support requested by `micronaut-projects/micronaut-http-benchmarks#3`.

## Changes proposed

- Add a built-in `handler.body.check` processor with exactly one of `equalTo`, `regex`, or `json`.
- Match complete fragmented response bodies and invalidate mismatched requests through the existing request lifecycle.
- Compile expected JSON into immutable recursive matchers and stream actual JSON through Jackson Core without materializing an actual-value tree.
- Generate schema/reference documentation and add YAML, unit, and chunked HTTP integration coverage.
- Define mathematical JSON number equality so forms such as `1`, `1.0`, and `1e0` compare equally.

## AI disclosure

The code in this pull request is AI-generated and human-reviewed.

## Check List

## Verification

- `mvn -pl core,http,test-suite -am test`
- `mvn -pl core -am -Dtest=CheckProcessorTest -Dsurefire.failIfNoSpecifiedTests=false test` (18 tests)
- `mvn -pl http -am -Dtest=ErrorRatioTest -Dsurefire.failIfNoSpecifiedTests=false test` (9 tests)
- `mvn -pl core,http -am process-sources`
- `mvn -pl distribution -am package -DskipTests`

Resolves micronaut-projects/micronaut-http-benchmarks#3


yawkat:
I saw there's some manual JSON parsing in other processors, but imo that is overkill for this purpose, it would lead to much more code. I used jackson-core instead, which is a new dependency, but I think that is also not an issue since it's already used elsewhere?

franz1981:
JDK 21 errors seems on a flacky test @diegolovison ?
Maybe we should add testlens or any similar mechanism to report on non-flacky tests only?

franz1981:
@yawkat 

Re 

> that is overkill for this purpose

AFAIK it was manually done for performance reason (and in order to act straight on Netty buffers and in an iterative/streaming way (with, I agree, an overkill complex state machine, sigh); that said the sentiment of being performance efficient, is still valid, so I'll talk with @diegolovison as we probably want this to be as GC free as possible and efficient before accepting a third party dep :pray: 
(unless you already have some profiling data of this check!)

yawkat:
Yea, there's some merit in making it low allocation, I just don't think it's worth the maintenance cost. In my use case, I will only occasionally check requests anyway, so it's not on the benchmark hot path. It's more of a sanity check.

diegolovison:
Fair enough. Let me just request a quick note in the class Javadoc and the docs explaining that this is for occasional checks rather than the hot path, so Jackson is preferred for simplicity over low-allocation optimizations. That will give great context for future maintainers who might wonder about performance optimizations.

PR #874, opened 2026-08-31
Add strict standalone run result handling

yawkat:
## Fixes Issue

Related to https://github.com/micronaut-projects/micronaut-http-benchmarks/issues/4

## Changes proposed

- Propagate the standalone `run.sh` command result to the process exit status.
- Add opt-in `--fail-on-errors` handling for runtime, validation, SLA, phase, cancellation, and incomplete-run failures.
- Add `--export DESTINATION` with optional `--export-format JSON|CSV` for final run statistics.
- Add focused standalone CLI tests and document the new options.

## AI usage

This pull request was developed with assistance from OpenAI GPT-5.6 through OpenCode. AI assistance was used to investigate the Hyperfoil and Aesh command paths, draft implementation and tests, and refine documentation. The resulting changes were manually reviewed and validated with the project test suite.


cli/src/main/java/io/hyperfoil/cli/commands/LoadAndRun.java
```diff
@@ -26,7 +29,7 @@ public static void main(String[] args) {
       boolean clustered = Arrays.asList(args).contains(CLUSTERED);
       args = Stream.of(args).filter(s -> !CLUSTERED.equals(s)).toArray(String[]::new);
       LoadAndRun lr = new LoadAndRun(clustered);
-      lr.exec(args);
+      System.exit(lr.exec(args));
```
franz1981, line 32: why this @yawkat ?

yawkat: without this, LoadAndRun prints an error but always exits with status 0. This change lets you use hyperfoil in CI to check that the load executed successfully.

imo there's no reason not to propagate the exit code. If people want to use this script programmatically they can use LoadAndRun.exec (like the tests already do) instead of calling main.

stalep:
Hi @yawkat , thanks for the pr!
I think propagating the exit codes and adding --fail-on-errors are useful to add, a few thoughts on the pr;

Could we centralize exit code handling in BaseStandaloneCommand? Also, the latest version of æsh (3.17.4) propagates other values as well depending on the situation;  [CommandResult](https://github.com/aeshell/aesh/blob/master/aesh/src/main/java/org/aesh/command/CommandResult.java)

Also, export path with spaces will most likely break I think, it would be better to use a Resource object as the Export command does. In addition if export fails the existing Export command catch IOException but returns SUCCESS. Combined with --fail-on-errors a user could get exit 0 even though the export file was not written.

yawkat:
Addressed the export feedback in 861c802e:

- `--export` is now parsed as an Aesh `Resource` and passed directly to `Export`, avoiding command-string re-parsing; the test uses a path containing spaces.
- `Export` now returns `CommandResult.FAILURE` on write errors, and `LoadAndRun` propagates that failure.

The focused standalone tests pass (9/9).

yawkat:
(above comment AI-generated, the commit too)

I don't think centralizing in BaseStandaloneCommand is a good idea. System.exit calls should be at the edge in the `main` methods only. That way, nobody is tempted to call these methods programmatically.

cli/src/main/java/io/hyperfoil/cli/commands/LoadAndRun.java
```diff
@@ -82,6 +97,55 @@ protected void monitor(HyperfoilCommandInvocation invocation) throws CommandExce
...
+            throw new CommandException("Export destination must not be empty");
+         }
+      }
+
+      private void failOnErrors(HyperfoilCommandInvocation invocation) throws CommandException {
+         Client.RunRef runRef = invocation.context().serverRun();
+         io.hyperfoil.controller.model.Run run = runRef.get();
+         RequestStatisticsResponse stats = runRef.statsTotal();
+         if (hasErrors(run, stats)) {
+            throw new CommandException("Run " + runRef.id() + " completed with errors");
```
yawkat: Throwing `CommandException` here (and on the export failure path at line 107) means `BaseStandaloneCommand.exec` handles it in its generic `catch (Exception e)` block, which prints `Failed to execute command: Run 0000 completed with errors` and then the **full usage help** for `run` (see `BaseStandaloneCommand.exec` lines 75-88). A benchmark that legitimately finished with SLA/validation errors ends up looking like the user mistyped the command line.

The root cause is that `Run.execute` unconditionally returns `SUCCESS` after `monitor()`, so a subclass has no non-exception way to report failure. Letting `monitor()` (or `Run.execute`) return a `CommandResult` and reserving the help print for parser/usage exceptions would fix both paths cleanly.

cli/src/main/java/io/hyperfoil/cli/commands/LoadAndRun.java
```diff
@@ -82,6 +97,55 @@ protected void monitor(HyperfoilCommandInvocation invocation) throws CommandExce
...
+         }
+      }
+
+      static boolean hasErrors(io.hyperfoil.controller.model.Run run, RequestStatisticsResponse stats) {
+         return run.cancelled || !run.completed || !run.errors.isEmpty() || run.phases.stream().anyMatch(phase -> phase.failed)
+               || stats.statistics.stream().anyMatch(LoadAndRunCommand::hasErrors);
+      }
+
+      private static boolean hasErrors(RequestStats stats) {
+         return !stats.failedSLAs.isEmpty() || stats.summary.invalid > 0 || stats.summary.requestTimeouts > 0
```
yawkat, line 143: This is applied to every entry from `statsTotal()`, including phases flagged `isWarmup`. A warmup phase that records a few connection errors or invalid responses before the server is fully ready will fail the whole run under `--fail-on-errors`, even when the measured phases are clean and have no SLA failures. `RequestStats` carries the `isWarmup` flag, so it's easy to skip those (or document that warmup errors count).

cli/src/main/java/io/hyperfoil/cli/commands/LoadAndRun.java
```diff
@@ -82,6 +97,55 @@ protected void monitor(HyperfoilCommandInvocation invocation) throws CommandExce
...
+               throw new CommandException("Interrupted while exporting run statistics", e);
+            }
+         }
+         if (failOnErrors) {
+            failOnErrors(invocation);
+         }
+      }
+
+      private void parseExport() throws CommandException {
+         if (export == null) {
```
yawkat: Because this returns early when `--export` is absent, `--export-format` is silently accepted and ignored on its own: `run.sh --export-format CSV bench.yml` (forgot `--export`) runs, produces no export, and exits 0, and `--export-format XML` without `--export` passes validation too. For an option aimed at unattended runs it'd be safer to validate the format unconditionally, or reject `--export-format` without `--export`.

yawkat (commented):
Review of the standalone run result handling. Findings are inline; the first one (failure paths printing the usage help) is the main thing I'd address before merging.

cli/src/main/java/io/hyperfoil/cli/commands/LoadAndRun.java
```diff
@@ -82,6 +97,55 @@ protected void monitor(HyperfoilCommandInvocation invocation) throws CommandExce
...
+            return;
+         }
+         exportFormat = exportFormat.toUpperCase();
+         if (!"JSON".equals(exportFormat) && !"CSV".equals(exportFormat)) {
```
yawkat: Minor: this re-implements the JSON/CSV check that `BaseExportCommand.getAcceptFormat()` / `getDefaultFilename()` already do (with a slightly different error text). Building the `Export` instance in `setup()` and calling `getAcceptFormat()` would give the same fail-fast behavior without a second copy of the accepted-format list.

cli/src/main/java/io/hyperfoil/cli/commands/LoadAndRun.java
```diff
@@ -26,7 +31,7 @@ public static void main(String[] args) {
       boolean clustered = Arrays.asList(args).contains(CLUSTERED);
       args = Stream.of(args).filter(s -> !CLUSTERED.equals(s)).toArray(String[]::new);
       LoadAndRun lr = new LoadAndRun(clustered);
-      lr.exec(args);
+      System.exit(lr.exec(args));
```
yawkat: Good change. `Wrk.main` and `Wrk2.main` follow the same `BaseStandaloneCommand` pattern and still discard the `exec` result, so `wrk.sh` / `wrk2.sh` keep exiting 0 on failure. Worth applying the same one-liner there while you're at it.

docs/site/content/en/docs/howtos/run_scripts.md
```diff
@@ -63,6 +68,15 @@ For instance:
 ./distribution/bin/run.sh -o /tmp/reports /tmp/first-benchmark.yml
 ```
 
+For unattended execution with isolated controller state and JSON results:
```
yawkat: Nothing in this example isolates controller state: `--fail-on-errors` and `--export` don't change where the in-VM controller keeps its data (`StartLocal` still uses the default `/tmp/hyperfoil` root). I'd drop "with isolated controller state" or show how to actually isolate the root dir.

test-suite/src/test/java/io/hyperfoil/benchmark/standalone/RunTest.java
```diff
@@ -1,50 +1,127 @@
...
 public class RunTest extends BaseBenchmarkTest {
 
-   @Test
-   public void testRunMain() {
-      String benchmark = getBenchmarkPath("scenarios/httpRequestParameterized.hf.yaml");
-      LoadAndRun.main(new String[] { "-PSERVER_PORT=" + httpServer.actualPort(), benchmark });
-   }
-
    @Test
    public void testRun() {
```
yawkat, line 31: With `testRunMain` removed (understandable, since `main` now calls `System.exit`), the `--clustered` filtering and the exit-status wiring in `LoadAndRun.main` have no coverage. Extracting a `static int run(String[] args)` that `main` wraps in `System.exit` would keep that path testable.

test-suite/src/test/java/io/hyperfoil/benchmark/standalone/RunTest.java
```diff
@@ -1,50 +1,127 @@
...
+
+      int result = new LoadAndRun(false).exec(new String[] { "--export", resultFile.toString(), "--export-format", "CSV",
+            "-PSERVER_PORT=" + httpServer.actualPort(), benchmark });
+
+      assertEquals(CommandResult.SUCCESS.getResultValue(), result);
+      byte[] content = Files.readAllBytes(resultFile);
+      assertTrue(content.length >= 2 && content[0] == 'P' && content[1] == 'K');
+   }
+
+   private String writeBenchmark(Path tempDir, String name, int durationSeconds, String expectedStatus) throws IOException {
```
yawkat: Nit: this helper has a single caller passing constants (`"invalid-response", 0, "5xx"`), so the `name`/`durationSeconds`/`expectedStatus` parameters only add indirection. Inlining the YAML in the test (or dropping the parameters) reads more directly.

cli/src/main/java/io/hyperfoil/cli/commands/LoadAndRun.java
yawkat: I don't want to touch wrk/2 and risk breaking users

